PR #2, opened 2021-07-23
[WIP] Use CSS Custom Properties for all Sass variables

clhenrick:
- allow for using Sass variables in the `.svelte` components as CSS custom properties
- allow for accessing the Sass variables in JavaScript as CSS custom properties
- fixes the ESLint configuration (should have been done in #8 but was missed)
- adds Sass and CSS vars for Carbon spacing in rem units
- adds an npm script for test building the Sass as it's faster than waiting on Sapper to do it.


clhenrick:
@mukhtyar in this PR we may also want to go through the existing Svelte components and replace the hard coded color variables in the `<style>` blocks with their equivalent CSS Custom properties. This will make them easier to track down in the future if we end up updating our Sass/CSS variables and/or Cal Adapt theme colors.

clhenrick:
We should probably add variables for [spacing](https://www.carbondesignsystem.com/guidelines/spacing/overview) as well. It supposedly can be imported from the [@carbon/layout](https://github.com/carbon-design-system/carbon/tree/main/packages/layout/scss) package which should be installed with the carbon components library.

src/routes/tools/climate-extremes/Header.svelte
```diff
@@ -23,7 +23,7 @@
     top: 0;
     border-bottom: 1px solid #dadee1;
     z-index: 3;
-    background: #9fa6ac;
+    background: #var(--gray-30);
```
mukhtyar: Do we need a `#` here?

src/routes/tools/annual-averages/Header.svelte
```diff
@@ -23,7 +23,7 @@
     top: 0;
     border-bottom: 1px solid #dadee1;
     z-index: 3;
-    background: #9fa6ac;
+    background: #var(--gray-30);
```
mukhtyar: Do we need a `#` here?

mukhtyar (approved):
Thanks for taking care of this!

src/routes/tools/climate-extremes/Header.svelte
clhenrick: fixed.

src/routes/tools/annual-averages/Header.svelte
clhenrick: fixed.

clhenrick:
@mukhtyar just FYI I also added Sass and CSS variables for the Carbon spacing units which are mostly in multiples of 4 and 8 pixels. We can use these in the future for things like padding, margin, grid-gap, etc. to help keep the layout consistent between pages and to not have to worry about doing pixel to rem calculations.

PR #4, opened 2021-07-23
[WIP] Getting Started pages

clhenrick:
Adds the "Getting Started" landing page and topic pages

PR #6, opened 2021-07-31
Add help/get-started pages

clhenrick:
This PR supersedes #4 as it does a better job of utilizing the existing patterns in `src/routes/help`.

- [x] Getting Started landing page
- [x] Getting Started topic pages
- [x] Adds topic pages to left sidebar navigation
- [ ] ~~Adds topic page anchors to right sidebar navigation~~
- [x] Adds previous and next topic links to bottom of topic pages


clhenrick:
[Help: Get Started](https://trello.com/c/2aKsnNzX/15-help-get-started)

clhenrick:
@mukhtyar folliwing up on our discussion today, these were the following things I'm curious about doing with the Getting Started topic pages and Sapper:

- indented links in the left side nav for the getting started topic pages
- page anchor links for headings in each getting started topic page
- previous and next links at the bottom of an article for moving between topic pages

But if we are planning on migrating to Svelte Kit soon then maybe it's not worth looking into too much right now? So please don't feel like you need to spend too much time on it or feel free to put it off till after we migrate. Thank you!

mukhtyar:
When you land on `/help/get-started/` you have access to 2 variables `toc` and `data`. Currently only the `toc` variable is passes in as prop to `_SidebarLeft.svelte` component. I think if you pass in the `data` as prop, you should have access to the list of items under `get-started`.

In `_SidebarLeft.svelte` if you uncomment the lines 4-5 and 22-34 and update the `if` condition you will see indented links under Get Started.

mukhtyar:
To add Previous and Next links at bottom of each page, update the `_ItemDetail.svelte` with new props and pass in the previous and next urls. This will add it to any content that uses `_ItemDetail.svelte`, at this point I think it's items under FAQ and Get Started.

mukhtyar:
We can skip page anchor links for headings in each getting started topic page for now. [mdsvex](https://mdsvex.com/docs) seems like a more powerful markdown parser, we can switch to that when we migrate to svelte-kit.

clhenrick:
thanks for the info @mukhtyar, that sounds good to me.

clhenrick:
@mukhtyar this one is ready for your review, no rush! 

Some brief notes:

- I added an alias in the webpack config for the `src/` directory, so instead of navigating up a path like `../../../dir` you can just do `~/dir`
- the `_getting-started-topics.js` contains data similar to `_categories.js` but I omitted the `dir` prop since these files will always come from `content/get-started/`
- I made the footer that has the support email contact its own component
- I added a few other components: Get Started, TopicTile, ItemNavPrevNext
- The `ItemNavPrevNext` component could also be used for the FAQs page if we want
- I tried to update the `SidebarLeft` so that it isn't hardcoded to `get-started` in case we want to use the nested TOC pattern for other sections in Help.

Let me know if you have any questions or suggestions, happy to jump on a quick Zoom call or talk over chat. Thanks!

src/partials/SidebarRight.svelte
```diff
@@ -1,11 +1,15 @@
 <script>
+  import { stores } from '@sapper/app';
   import { createEventDispatcher } from 'svelte';
   import { Search, Select, SelectItem } from "carbon-components-svelte";
 
   export let display = ['events', 'posts'];
   export let events = [];
   export let posts = [];
   export let filters = [];
+  export let anchors = [];
```
mukhtyar, line 10: So any sub-headings will show up in the right sidebar? I like that idea.

src/routes/help/[category]/[slug].svelte
```diff
@@ -97,13 +122,15 @@
     <div class="bx--grid">
       <!-- Row -->
       <ItemDetail {item} />
+      <Row>
+        <Column>
+          <ItemsPrevNextNav prev={prevArticle} next={nextArticle} />
```
Comment (mukhtyar) [inline src/routes/help/[category]/[slug].svelte:127]: Adding it to FAQs might be useful too.

src/routes/help/_TopicTile.svelte
```diff
@@ -0,0 +1,87 @@
+<style lang="scss">
```
mukhtyar, line 1: Just realized the prettier config has `svelteSortOrder: "styles-markup-scripts"` and I've been doing `"scripts-styles-markup"` order. We should probably decide on one.

mukhtyar (approved):
Looks great! We should talk about some common eslint rules :-)

src/partials/SidebarRight.svelte
clhenrick: Thanks @mukhtyar and yes, that was my thinking. It could be helpful for longer articles with many headings > h1 but we might want to disable it for articles that have only a single h2 heading.

src/routes/help/_TopicTile.svelte
clhenrick: Ah that's why that's happening! I thought it was just my VS Code settings so was trying not re-format any existing files. I'm not partial to either, perhaps it makes sense to use `"scripts-styles-markup"` as that's the pattern used through the majority of the Svelte files.

clhenrick:
Thanks @mukhtyar for the review! I'll add a Trello card for us to follow up on the ES Lint & Prettier configurations and one for adding the previous & next article links to the FAQs page.

PR #9, opened 2021-08-16
Extreme weather

### LINKED ISSUE #7: Clearing model selection after fetching data breaks app
On the Annual Averages tool, if I fetch data for a location then clear the existing selected models, the UI enters a loading state which it doesn't recover from, perhaps from a network request failure to Cal Adapt API:

image

mukhtyar:
This pull request contains updates to Annual Averages tool and Extreme Weather tool and several shared components.
- Moves location selection to a modal
- Updates static map styling
- Use new layout as we discussed
- Cleans up and adds more preload functionality like retrieving datasets descriptions from API
- Cleans up the geocode functions (still need to remove old ones used in other tools)


src/components/tools/Charts/Histogram/Annotation.svelte
```diff
@@ -1,29 +1,74 @@
 <script>
   import { getContext } from "svelte";
 
-  const { padding, yScale, xScale } = getContext("LayerCake");
+  const { yScale, xScale } = getContext("LayerCake");
 
   export let data;
+
+  $: percentiles = data.percentiles;
```
clhenrick, line 9: More of a Svelte question, does referencing object properties using the `$` label make these variables reactive when normal object de-structuring would not? For example if instead you did `const { percentiles, low, high } = data` would Svelte not react to updates?

src/components/tools/Charts/Histogram/Chart.svelte
```diff
@@ -1,7 +1,7 @@
 <script>
   import { SkeletonText, SkeletonPlaceholder } from "carbon-components-svelte";
   import { LayerCake, Svg, Html } from "layercake";
-  import { histogram, extent } from "d3-array";
+  import { bin, thresholdFreedmanDiaconis, extent } from "d3-array";
```
clhenrick: I did not know about `thresholdFreedmanDiaconis`, neat!

src/components/tools/Charts/Histogram/Chart.svelte
```diff
@@ -11,7 +11,7 @@
 
   export let data;
   export let labels;
-  export let height = "350px";
+  export let height = "400px";
```
clhenrick: Small suggestion, I'm not sure if this is how it's done for all Charts, but making this prop a number might be less confusing when using it elsewhere. You could then have a reactive prop that's derived from `height` like so:

```
$heightPx: `${height}px`;
```

src/helpers/geocode.js
```diff
@@ -112,6 +115,39 @@ export const createAdditionalProps = (feature, layerId) => {
   return { title, address };
 };
 
+export const getTitle = (feature, layerId, placeName) => {
+  let title;
+  switch (layerId) {
```
clhenrick, line 119: this way is totally fine but you could also just `return` the string value for each statement, that way you wouldn't need the local `title` variable or the `break`.

src/helpers/utilities.js
```diff
@@ -142,7 +142,7 @@ export const handle = (promise) => {
  * @param {object} params - query parameters
  * @return {Promise}
  */
-export function fetchData(url, params, method = "GET") {
+export function fetchData(url, params = {}, method = "GET") {
```
clhenrick, line 145: Because Sapper has it's own internal `this.fetch` it might be good to use `window.fetch` in this function to help distinguish the two versions of `fetch`. Generally I prefer to include `window` when using Browser APIs such as fetch, setTimeout, etc. so that it reinforces that it's a global, but just my personal preference.

clhenrick (approved):
I did a cursory look over the code, I didn't want to go too deep knowing that you're still working on the Extreme Weather tool. 

Some thoughts / questions:

- Can we abstract the commonly used components of the tools pages like Header, About, Help, Resources, Data, etc. so that each tool doesn't have to contain its own version of them? Maybe there could be a directory `src/routes/tools/common/` that contains these common components and each tool could import them. It would make the tools easier to maintain moving forward IMO.

- For the tool description in the header, let's set the descriptive text below the title to `max-width: min(70ch, 70vw)` to keep the text from spanning the entire width of the page.

- For the sticky nav, is there a reason we don't include all of the various sections in it and only have a few of them? It seems a little odd to leave out "Data Sources" and "Related Tools".

- For the headings below the dashboard, we could reduce the vertical space between "Help" (h2) and "About the Tool" (h3). We could also make the h3 headings a slightly smaller font size (e.g. 24px) to convey that they're sub-headings of the Help section. Or, if there's not going to be any immediate text under "Help" then we could remove Help and make each of the headings h2s.

- Noting that it would be good to run the Prettier formatting npm script so that future changes to these files don't include code formatting in the git history.

clhenrick:
@mukhtyar one other thing I just noticed with the static map style change is that the boundaries can be a little difficult to read, maybe try using a 2px or 3px stroke:

Screen Shot 2021-08-16 at 5 10 04 PM

Screen Shot 2021-08-16 at 5 09 39 PM

They also get a little too simplified at times, maybe reduce the simplification algorithm:

- Census tract that shouldn't be a triangle:
Screen Shot 2021-08-16 at 5 12 05 PM

- Marin county looks a little too jagged:
Screen Shot 2021-08-16 at 5 13 15 PM

I just tried adding a slight opacity (~60-70%) to the static map, it helps it not stand out too much since it's the only thing with a lot of color in the dashboard now:
Screen Shot 2021-08-16 at 5 17 56 PM

src/components/tools/Charts/Histogram/Chart.svelte
mukhtyar: I didn't till recently either :-), just played around with some binning algos in d3, this "looked" the best. Having number of bins same for all locations made the histogram weird.

src/components/tools/Charts/Histogram/Chart.svelte
mukhtyar: Good idea!

mukhtyar:
The Static Maps API has a limit of length of requests, so I've simplified the geojson overlays I include. Will play around with the simplification settings to see how far we can go.

src/components/tools/Charts/Histogram/Annotation.svelte
mukhtyar: You can't use `let`, `const` on a reactive line in svelte. I think if I wrap that in brackets, it might work.

mukhtyar: I meant the Data Sources and Related Tools to be sub-headings of About and Resources. I will update the heading styles. Open to making all of them h2s.

clhenrick:
[Clearing model selection in settings breaks the Annual Averages tool](https://trello.com/c/vP2az7mE/58-clearing-model-selection-in-settings-breaks-the-annual-averages-tool)

PR #11, opened 2021-08-20
Annual averages

mukhtyar:
Move the About, Help and Resources and other sections common to all tools to a new Partials dir. Loads the ChangeLocation and DownloadChart components asynchronously. Also fixes  the oversimplification bug in StaticMaps.


src/components/tools/Partials/Header.svelte
```diff
@@ -0,0 +1,28 @@
...
+  export let iconPaths;
+</script>
+
+<!-- Banner -->
+<div class="banner bg-gray-100">
+  <div class="bx--grid">
+    <div class="bx--row">
+      <div class="bx--col">
+        <div class="center-row">
+          {#if iconPaths && iconPaths.length > 0}
```
clhenrick, line 11: Minor suggestion, feel free to ignore: for array checking I think it's good to do `Array.isArray(someArray) && someArray.length` to be a little more certain that it is an array.

src/components/tools/Partials/Help.svelte
```diff
@@ -0,0 +1,38 @@
...
+  <div class="bx--col-lg-12">
+    <h2>Help</h2>
+    <slot name="help">
+      {#if !items || items.length === 0}
+        <em>[Provide a list of Help items for the tool]</em>
+      {:else}
+        {#each items as opt}
+          <div class="bx--row">
+            <div class="bx--col">
+              <ClickableTile href="{opt.link}" class="center-row">
```
clhenrick: Is `<ClickableTile>` one that we'd want to use our own custom `<Card>` for?

clhenrick (approved):
Looks good to me. I like the fallbacks in the `<slot>`s to remind folks to write copy for the tools later.

My only suggestions are for improving the typography outside of the dashboard:
- For the text below the title in the header, `max-width: 60ch` looks good
- For the Help section paragraphs we could bump up the font-size to 20px / 1.5rem, set the `max-width: 75ch`, and `margin-bottom: 1.75rem`, also for the list items.

We could also improve the alignment between the dashboard and Help section by giving both the same `margin-left` for now, it looks a bit odd on larger desktop displays:

Screen Shot 2021-08-20 at 5 11 06 PM

clhenrick:
Also, we don't need to fix this immediately, but I'm wondering if we could make the "F˚" in the stats box a hanging punctuation. 

Instead of:

Screen Shot 2021-08-20 at 5 14 53 PM

maybe do this?

Screen Shot 2021-08-20 at 5 17 22 PM

src/routes/tools/annual-averages/ExploreData.svelte
```diff
@@ -54,14 +55,38 @@
...
+  let DownloadChart;
+
+  let metadata;
+  let csvData;
+  let printContainer;
+  let printSkipElements;
+
+  function loadLocation() {
+    showChangeLocation = true;
+    import("../../../components/tools/Partials/ChangeLocation.svelte").then(
```
clhenrick: I think you can also use `async/await` here:

```js
  async function loadLocation() {
    showChangeLocation = true;
    ChangeLocation = (
      await import("../../../components/tools/Partials/ChangeLocation.svelte")
    ).default;
  }
```

src/routes/tools/annual-averages/ExploreData.svelte
```diff
@@ -54,14 +55,38 @@
...
+      ["boundary", $boundary.id],
+      ["feature", $location.title],
+      ["center", `${$location.center[0]}, ${$location.center[1]}`],
+      ["scenario", $scenario.label],
+      ["climate indicator", $climvar.label],
+      ["units", $climvar.units.imperial],
+    ];
+    printContainer = document.querySelector(".explore");
+    printSkipElements = ["explore-settings"];
+    import("../../../components/tools/Partials/DownloadChart.svelte").then(
```
clhenrick: ditto here too:

```js
    DownloadChart = (
      await import("../../../components/tools/Partials/DownloadChart.svelte")
    ).default;
```

src/routes/tools/annual-averages/ExploreData.svelte
mukhtyar: Changed the units to a hanging punctuation. This format did not work for earlier iterations because the box had > 2 values, and longer units like `inches` & values with more significant digits. Will need to keep an eye on it for future tools.

clhenrick: That makes sense, thanks for pointing that out to me.

PR #14, opened 2021-08-26
Data download page

clhenrick:
- Adds content, styles, and links for the top level data download page
- Adds some helper CSS classes for adjusting spacing
- Updates typography styles for h1-h6 and p elements
- Uses new layout via Carbon Grid for vertically aligning page elements

### DIFF
--- a/src/routes/data/index.svelte
+++ b/src/routes/data/index.svelte
@@ -23,7 +23,8 @@
 <div class="bleed banner overlay">
   <div class="bx--grid">
     <div class="bx--row">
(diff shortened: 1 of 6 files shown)


clhenrick:
@mukhtyar I also updated some of the default font styles (h1-h6, p, p.lead). The way `@includes` was being used in places was overriding other CSS rules for the selector. With `@includes` I believe you want it to be the first thing in an element's style rules, that way any other style rules override it and not vice versa.

mukhtyar (approved):

PR #16, opened 2021-08-26
Remove developers page

clhenrick:
- [x] remove `src/routes/developer`
- [x] remove "developer" from main navigation
- [x] update homepage card with "developer" to link to `/data/#cal-adapt-api`
- [x] add page anchors to `h2` headings on Data page
- [x] add sticky navigation to Data page for headings
- [x] update dummy event so it shows up on homepage


clhenrick:
@mukhtyar the only thing I'm unsure about is what should take the place of the "Developer" Card in the homepage. Perhaps "Events"? But I think that also gets shown in the right sidebar (but only when there's an upcoming event). Let me know if you have any ideas?

mukhtyar:
We can add a link to the Cal-Adapt API section on the Data page for the Developers card on the homepage.

clhenrick:
That's a great idea, thanks. That means we’ll need page anchors to each
section on the data page. Maybe adding the sticky nav similar to other
pages like with tools would be helpful as well?

mukhtyar (approved):

mukhtyar:
The Developers card in `content/homepage/cards.js` can point to the Cal-Adapt API section on the Data page

mukhtyar:
Yes, adding a sticky nav to the Data page is a great idea! Also can you add the upcoming Analytics Engine workshop as a new event, details [here](https://cal-adapt.org/blog/2021/analytics-engine-workshop/)?

clhenrick:
@mukhtyar should that event belong on the current/old site? Or are we going to point people to the unfinished dev site?

mukhtyar:
It's on the current site. My thought on adding it to dev site was that it might help designing a top level Events page. Optionally you can just change the date for the dummy event we currently have. Either would work. We are not pointing people to it.

clhenrick:
That makes sense. I would prefer to open a separate PR for the top level Events page but I'll update the dummy event for sometime in the year 2100.

PR #18, opened 2021-08-27
Annual averages fix bookmark

mukhtyar:
This pull request fixes units for precipitation envelope, moves bookmark to load async, and adds default help items.


src/components/tools/Partials/Help.svelte
```diff
@@ -1,38 +1,32 @@
...
-<!-- Resources -->
 <div class="bx--grid">
   <div class="bx--row">
-    <div class="bx--col-lg-12">
+    <div class="bx--col">
       <h2>Help</h2>
       <slot name="help">
-        {#if Array.isArray(items) || items.length === 0}
-          <em>[Provide a list of Help items for the tool]</em>
+        {#if !items || items.length === 0}
```
clhenrick: you could simply this if/else block by reversing the condition:

```svelte
        {#if Array.isArray(items) && items.length}
          <CardsContainer cardWidth="{cardWidth}">
            {#each items as { title, text, slug }}
              <Card
                {...{
                  titleText: title,
                  description: text,
                  linkPath: `/help/${slug}`,
                  ctaText: "Learn More",
                }}
              />
            {/each}
          </CardsContainer>
        {:else}
          <em>[Provide a list of related Help items]</em>
        {/if}
```

Generally it's less cognitive overload to have the truthy part first then fallback to the falsey part with if/else.

src/components/tools/Partials/Resources.svelte
```diff
@@ -40,12 +40,12 @@
...
   <div class="bx--row">
     <div class="bx--col">
-      <h2>Related Cal-Adapt Tools</h2>
+      <h3>Related Cal-Adapt Tools</h3>
       <slot name="tools">
         {#if !resources || tools.length === 0}
           <em>[Provide a list of related Cal-Adapt tools]</em>
         {:else}
-          <CardsContainer gridGap="{2}" cardWidth="{toolCardWidth}">
+          <CardsContainer cardWidth="{toolCardWidth}">
```
clhenrick: For each `<Card>` you can set the heading element via the `headingLevel` prop which takes a number (e.g. 2 - 6).

src/components/tools/Partials/ShareLink.svelte
```diff
@@ -0,0 +1,26 @@
...
+  $: bookmark = `${pathname}?${state}`;
+</script>
+
+<Modal
+  id="share"
+  passiveModal
+  bind:open
+  modalHeading="Share Link"
+  on:open
+  on:close
```
clhenrick: This works (and it has keyboard a11y built in, yay!) but I don't see where `close` is defined? I'm probably missing something here.

src/components/tools/Settings/UploadBoundary.svelte
```diff
@@ -27,8 +28,13 @@
     fileUploader.clearFiles();
   }
 
-  function clear() {
-    console.log("clear");
+  function clearFiles() {
```
clhenrick, line 31: I did not know you could `bind` to a component instance and set properties or call methods on it like this, neat!

src/routes/tools/annual-averages/_store.js
```diff
@@ -184,11 +184,9 @@ export const bookmark = derived(
...
     const [lng, lat] = location.center;
-    const bookmark = `climvar=${$climvarStore}&scenario=${$scenarioStore}&models=${$modelsStore}
-    &imperial=${imperial}&lng=${lng}&lat=${lat}&boundary=${boundaryId}`;
-    if (process.browser) {
-      return `${window.location.href}?${bookmark}`;
+    if (boundaryId === "custom") {
+      return "Cannot create a bookmark for an uploaded boundary";
     }
-    return null;
+    return `climvar=${$climvarStore}&scenario=${$scenarioStore}&models=${$modelsStore}&imperial=${imperial}&lng=${lng}&lat=${lat}&boundary=${boundaryId}`;
```
clhenrick, line 190: Would it make sense to cache the address in here as well? When change the location to Oakland then open the share URL in a new browser tab it gives me a different address:

Screen Shot 2021-08-27 at 3 49 05 PM

Screen Shot 2021-08-27 at 3 49 00 PM

clhenrick (changes requested):
@mukhtyar in addition to the comments I left I think that it would be good to fix the use of heading elements (h1 - h6) here since we're templating this page for a lot of the other tools. 

The large bolded text in the Set Location UI shouldn't be an `<h3>` element as there is no `<h2>` preceding it. Alternatively I believe that there's an `.h3` CSS class that will achieve the same look: 
image

The text for "Change Settings" should not be an `<h4>` element because there's no `<h2>` or `<h3>` that precedes it. Probably best to just style the type using CSS.

IMO it makes more sense for the Data Sources and Related Tools to be `<h2>`s as they could be considered their own sections instead of sub-sections. It would also make it more consistent with Resources and Help. But up to you, not a big deal either way.

All the titles in the Cards at the bottom are currently `<h2>`s which isn't semantically correct. You can pass a prop to each card specifying what level the heading should be (it won't have an effect on the size of the text). If they're below an `<h2>` element then they should be `<h3>`s; if they're below an `<h3>` element then they should be an `<h4>`, etc.

src/components/tools/Partials/Help.svelte
mukhtyar: Made heading levels in all partials `h2` & switched order of if statement via https://github.com/berkeley-gif/caladapt-website-2021/pull/18/commits/6e0fc538ac1fbe949b1725b48659b54c35a7c502 and https://github.com/berkeley-gif/caladapt-website-2021/pull/18/commits/f7e8dd20675b0922c8ef414b50352ffab8d898bf

src/components/tools/Partials/Resources.svelte
mukhtyar: That's cool! Added `headingLevel` to the card components

src/components/tools/Partials/ShareLink.svelte
mukhtyar: Removed the unused events.

src/routes/tools/annual-averages/_store.js
mukhtyar: It would be good to make this change, but I need to think about the best way to do this. Currently when a tool loads, I use reverse gecoding since I only have the lat, lng. When you search for a location and update chart, I use forward geocoding. So the results returned from the 2 different MapBox endpoints are different. This is an issue only for LOCA grid cells, other boundaries wont have this issue since the names are from the Cal-Adapt API.

clhenrick: Ah okay, that makes sense. Seems like an edge case then but a potentially confusing one for users. We can log it in Trello and come back to it later if you don't want to tackle it now.

mukhtyar: Made a note of the [issue](https://trello.com/c/i6nQgTF3/66-make-location-name-retrieved-from-mapbox-consistent-for-the-same-grid-cell) in Trello.

src/components/tools/Partials/About.svelte
```diff
@@ -29,85 +29,79 @@
...
+      </p>
+      {#if !datasets || !datasetList}
+        <em>[Provide a list of datasets used for the tool]</em>
+      {:else}
+        {#await datasetList}
+          <InlineLoading description="Loading list of datasets..." />
+        {:then items}
+          {#each items as item}
+            <div class="bx--row source">
+              <div class="bx--col-lg-2">
```
clhenrick, line 59: When the viewport width narrows to 1056px and the grid switches from "large" to "medium"  (see [this list of breakpoints for the Carbon Grid](https://www.carbondesignsystem.com/guidelines/2x-grid/implementation#responsive-options_)) these two columns get stacked and the result is that the logo image becomes huge and blurry. We could fix this either by 1. not displaying the image at smaller screen widths, or 2. adding CSS classes with `bx--col-md-` to shift the layout (e.g. 1 column for the image and 7 columns for the text (there are 8 columns in the medium size grid)).

clhenrick: Maybe for now we can just add the medium column css classes since we already agreed we will be coming back to fixing the layout for mobile devices later?

src/routes/tools/annual-averages/_store.js
clhenrick: Thank you!

clhenrick (approved):
This is looking good, I like how the various tools sections are componentized now to make the code for the tools more dry. I'm wondering if at some point though we should move the CSS from `_tools.scss` into the various partials? It would probably help for maintainability and make it easier to edit / change (because Sapper is faster at performing SCSS updates when they are in `.svelte` files instead of `.scss` files)

src/components/tools/Partials/About.svelte
mukhtyar: Sounds good. I'll add the `bx--col-md-` classes and move the scss to partials in the next PR.

PR #20, opened 2021-09-01
Make events a top level page

clhenrick:
- [x] adds a new events top level page at `/events/`
- [x] removes existing events page & slugs at `/blog/events/` & `/blog/events/[slug]`
- [x] adds individual events pages at `/events/[slug]`
- [x] adds "events" to top level / main navigation
- [x] add subscribe button & text
- [ ] ~~adds pagination for past events~~ (punt this till later because we only have 9 past events to list)
- [x] fixes some a11y issues with events list (headings, use list items for events lists)
- [x] adds a TweetButton component that composes a Twitter "tweet" when clicked on


clhenrick:
[Events](https://trello.com/c/A7LD4MXj/57-events)

clhenrick:
@mukhtyar here is a [deploy preview](https://6137a0e80fc8d50925d0dd66--happy-engelbart-4042ee.netlify.app/events/) for the events top level page.  I decided to punt pagination as we currently only have 10 total events to display. I also added a new TweetButton component that creates a pre-populated tweet when clicked on.

src/components/social-media/TweetButton.svelte
```diff
@@ -0,0 +1,24 @@
...
+    data-dnt="true"
+    data-text="{text}"
+    data-url="{url}"
+    data-hashtags="{Array.isArray(hashtags) && hashtags.length
+      ? hashtags.join(',')
+      : ''}"
+    data-via="{via}"
+    data-size="{size}"
+  >
+    Tweet</a
```
mukhtyar: Can we add the Twitter icon here? Using the `Icon Button` or `Link with Icon` component

mukhtyar (approved):

src/components/social-media/TweetButton.svelte
clhenrick: @mukhtyar it should look like this:
image

but for some reason the Twitter JS isn't working as expected so it just shows the word "Tweet" which I think is what you're seeing? If you try refreshing the page then you should see the Tweet button with the Twitter logo. 

I'll dig into it and see if I make it work, I might have to put the Twitter JS code in the component's `onMount` lifecycle.

mukhtyar: Got it. Still see just the word `Tweet` after refreshing the preview and not the button.

clhenrick: I ended up rewriting the component and ditching the Twitter JS which was supposed to style the button. I think it looks better now if you want to check it locally (I haven't updated the Netlify preview deploy).

PR #21, opened 2021-09-07
Extreme weather cec updates

mukhtyar:
This pull request adds all the changes requested by the CEC, streamlines some of the charting functionality and creates a new Dashboard component using the Carbon grids that can be used by other tools.


src/components/tools/Charts/Histogram/Annotation.svelte
```diff
@@ -3,27 +3,24 @@
...
-
-  $: percentiles = data.percentiles;
-  $: low = data.low;
-  $: high = data.high;
+  export let units;
 </script>
 
+<style>
+  .title {
+    font-size: 18px;
```
clhenrick: Is there a reason why we should use pixels instead of rems for chart text? Like does the chart download not look correct when specifying this as `1.125rem` instead of `18px`? Just curious...

src/components/tools/Charts/Observations/AxisY.svelte
```diff
@@ -0,0 +1,77 @@
...
+</style>
+
+<g class="axis y-axis">
+  {#each tickVals as tick, i}
+    <g
+      class="tick tick-{tick}"
+      transform="translate({$xRange[0] +
+        (isBandwidth ? $padding.left : 0)}, {$yScale(tick)})"
+    >
+      {#if gridlines !== false}
```
clhenrick: You could simplify both these `#if` statements for gridlines and tickMarks by leaving out the `!== false` and `=== true`

src/components/tools/Charts/Observations/Chart.svelte
```diff
@@ -0,0 +1,87 @@
...
+        <AxisY gridlines="{true}" />
+        <Circles
+          y1="{-y1}"
+          units="{units}"
+          on:mousemove="{(event) => (evt = hideTooltip = event)}"
+          on:mouseout="{() => (hideTooltip = true)}"
+        />
+      </Svg>
+      <Html pointerEvents="{false}">
+        {#if hideTooltip !== true}
```
clhenrick: likewise you could simplify this to `{#if !hideTooltip}`

src/components/tools/Partials/About.svelte
```diff
@@ -56,16 +86,16 @@
         {:then items}
           {#each items as item}
             <div class="bx--row source">
-              <div class="bx--col-lg-2">
+              <div class="bx--col-lg-2 bx--col-md-1 bx--col-sm-1">
```
clhenrick, line 89: Nice, this does a good job of making the icon image responsive.

src/scss/site/_dashboard.scss
```diff
@@ -0,0 +1,94 @@
+#explore {
```
clhenrick, line 1: Is there a reason why these styles should live in a Sass partial instead of the `Dashboard.svelte` component? Or are these styles used elsewhere besides Tools?

clhenrick (approved):
@mukhtyar this is looking good, the dashboard layout responds really nicely to changing viewport widths. The way you implemented the Carbon Grid system in the dashboard isn't quite the way I was thinking of doing it, but it works. Happy to chat about it more this week, probably easier to do over a call then in a PR review.

src/scss/site/_dashboard.scss
mukhtyar: I can move the `#explore` styles inside `extreme-weather/index.svelte`. If I move these into `Dashboard.svelte` , each rule will need a `:global(...)` modifier. And also styles that are in `:global(...)` blocks cannot have any nesting like scss styles.

clhenrick: Got it. One other thought @mukhtyar is if these colors correspond to the climate models, should we have them stored as Sass & CSS variables in the `_variables.scss` partial so they can be referenced elsewhere?

mukhtyar: Yes that's a good idea. I'll add them in.

src/components/tools/Charts/Histogram/Annotation.svelte
mukhtyar: Updated to using rem in https://github.com/berkeley-gif/caladapt-website-2021/pull/21/commits/42fbbee541cccf1ba568c9556513b1b4016e6cab

src/components/tools/Charts/Observations/AxisY.svelte
mukhtyar: Thanks, fixed these in various places.

mukhtyar:
[Safe guards for NWS API flakiness](https://trello.com/c/gASTzuCW/9-safe-guards-for-nws-api-flakiness)

mukhtyar:
[Browser navigation broken on Extreme Weather tool](https://trello.com/c/zwhfooYH/30-browser-navigation-broken-on-extreme-weather-tool)

PR #22, opened 2021-09-07
Annual averages use dashboard

mukhtyar:
This pull request updates the Annual Averages tool to use the Dashboard component

### DIFF
--- a/src/routes/tools/annual-averages/ExploreData.svelte
+++ b/src/routes/tools/annual-averages/ExploreData.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { Button, Accordion, AccordionItem } from "carbon-components-svelte";
+  import { Button, Loading } from "carbon-components-svelte";
   import { format } from "d3-format";
   import { Download16, Share16, Location16 } from "carbon-icons-svelte";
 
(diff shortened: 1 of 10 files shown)


src/routes/tools/annual-averages/ExploreData.svelte
```diff
@@ -126,65 +127,119 @@
...
-      <div class="h3 block-title">{$location.title}</div>
-      <div class="h4">
-        Projected changes in <span class="block-title">{$climvar.title}</span>
-        under a <span class="block-title">{$scenario.labelLong}</span>.
-      </div>
-      <Button size="small" icon="{Location16}" on:click="{loadLocation}">
-        Change Location
-      </Button>
+<style lang="scss">
+  .block {
```
clhenrick, line 131: Question: won't some of this CSS be duplicated across tools? e.g. for `.block` and `.stats` since each tool has those components?

clhenrick (approved):
Looks good, I like how the various components of the Tools are coming together and utilizing the slots to help keep things dry. I wonder if it would make sense to componentize some of the Dashboard's child elements like the Title area, Stats, and Settings? That way the corresponding CSS could be reused with each of them. I suppose we could always refactor it later too.

src/routes/tools/annual-averages/ExploreData.svelte
mukhtyar: Yes, I think breaking up the Dashboard and moving the relevant css there would be a good next step.

PR #24, opened 2021-09-09
Homepage updates

clhenrick:
A11y fixes and styling updates for the homepage and some of its components:
- SidebarRight
- Footer


clhenrick:
[Homepage](https://trello.com/c/2qDQMH8H/1-homepage)

clhenrick:
@mukhtyar please take a look at it locally and let me know what you think, thanks! I didn't get around to addressing all of the vertical alignment issues between the various sections, it will take a bit more work so will come back to it later.

src/routes/index.svelte
```diff
@@ -49,43 +49,95 @@
...
 
 <!-- Banner -->
 <section
-  class="banner overlay overlay-black overlay-40"
+  class="bleed banner overlay overlay-black overlay-40"
   style="background-image: url(/img/banners/yosemite_2000w.jpg);background-size:cover;"
 >
   <div class="bx--grid">
-    <div class="bx--row" style="padding: 0.5rem 1rem;">
+    <div class="bx--row banner--icons" aria-hidden="true">
```
mukhtyar, line 111: Do we want to add a single anchor link at the row level which goes to the Tool page instead?

webpack.config.js
```diff
@@ -9,9 +9,10 @@ const mode = process.env.NODE_ENV;
 const dev = mode === "development";
 
 const alias = {
+  "~": path.resolve(__dirname, "src/"),
```
mukhtyar, line 12: These shortcuts are great!

mukhtyar (approved):

src/routes/index.svelte
clhenrick: My gut feeling is that I question the necessity of making these icons go to the tools page because we already link to that page twice; once in the navigation and another in the cards. If they were to link to the tools page then I think what would be ideal would be that they apply the filter associated with the icon to the list of tools.

webpack.config.js
clhenrick: Thanks! Yeah one benefit of using webpack's alias so you don't have to go crazy with `../`'s

PR #27, opened 2021-09-10
Add tool: degree days

clhenrick:
Implements the Cooling and Heating Degree Days tool

PR #29, opened 2021-09-13
Add tool for degree days

clhenrick:
Replaces #27  and uses the code from `src/routes/annual-averages/` via 4abf2ca as a starting point.


clhenrick:
[Tools: CDD/HDD](https://trello.com/c/CAGsPasI/50-tools-cdd-hdd)

content/tools/data.js
```diff
@@ -129,6 +129,31 @@ const tools = [
     icons: [
       'img/icons/sun.svg',
     ],
+    datasets: [
+      // Q: what do these refer to?
+      // TODO: probably only one or two relates to CDD/HDD...
+      {
+        slug: 'cb0c5f49-59d2-4f8e-8d9c-a047833f6081',
```
clhenrick, line 134: @mukhtyar what do these cryptic ids correspond to? Something in the API?

mukhtyar: Yes it's an identifier for the Dataset Instance in the API e.g. https://api.cal-adapt.org/api/datasets/cb0c5f49-59d2-4f8e-8d9c-a047833f6081/

src/routes/tools/degree-days/_data.js
```diff
@@ -0,0 +1,220 @@
...
+    return { ...d, slugs, mark: "line", visible: true, indicatorId };
+  });
+};
+
+// The ensemble has to be assembled from the max and min of 10/all models
+// Similar to the models, the models-max and models-min are 2 raster series each
+const getEnsembleSeries = ({ climvarId, scenarioId }) => {
+  return ENSEMBLES.filter((d) => d.id === `${scenarioId}_range`).map((d) => {
+    const slugs = [
+      `${climvarId}_year_models-min_historical`,
```
clhenrick: @mukhtyar I tried changing these to one listed in the google doc but it only returned a list of rasters, not actual data. I couldn't find more info the API docs, do you know what the correct API call would be? For example:

```
https://api.cal-adapt.org/api/series/cdd_year_ens32avg_historical/?imperial=true&g=Point(-121.4688%2038.5938)
```

clhenrick: Oh actually nevermind, I see that appending `/events/` to the URL path returns data.

mukhtyar: Yes. Although I just realized that these are pre-calculated CDD/HDD values with reference temperature set to 65. So if the user selects a different reference temp, these numbers would not update. In that case, we can just stick with the observed & models and not add the ensembles. Let's confirm this at our meeting today.

clhenrick: Got it, good catch

src/components/tools/Settings/SelectThresholdNumeric.svelte
```diff
@@ -0,0 +1,38 @@
+<script>
+  import { onMount, createEventDispatcher } from "svelte";
+  import { NumberInput, NumberInputSkeleton } from "carbon-components-svelte";
+  import { debounce } from "~/helpers/utilities";
+
+  export let value = 0;
+  export let title = "Select Threshold";
+
+  // TODO: these probably shouldn't be Infinity...
+  export let minValue = -Infinity;
```
clhenrick, line 8: @mukhtyar any guidance on what the minimum and maximum temperature thresholds should be for HDD/CDD? I won't hard code them here but noticed if the threshold value is too low or too high the API won't return data.

mukhtyar: While 65 is standard and useful if you are comparing climates, I haven't come across any literature on what the range could be. More at [this link](https://www.degreedays.net/introduction#base-temperature-confusion). If we talk about heating & cooling buildings, I'd guess something like 40-100 are good thresholds (although extreme). Maybe instead of defining thresholds, we can show a fallback No Data message.

src/routes/tools/degree-days/_data.js
bkg: If you'd like to add ensembles for CDD and HDD, you can compute them dynamically from the various tasmax 32 GCM ensembles with varying thresholds and frequencies like:

https://api.cal-adapt.org/api/series/tasmax_day_ens32avg_rcp85/cdd/?freq=A&thresh=65&g=Point(-121.4688+38.59)

clhenrick: Thanks @bkg that's helpful to know. 

In a somewhat related question, for the Cal-Adapt API is there a way to request specific months for the `freq` param? For example something like `freq=[5,6,7]` for grabbing only the months of June, July, and August. 

Currently in the dashboard we're filtering and then re-aggregating clientside; but from my basic stats background I don't think it's a good idea to do an average of averages, in this case averaging the filtered months from the API response which I assume themselves are already averages.

bkg: The `freq` param is for temporal aggregation and it supports any of the [Pandas offsets](http://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#dateoffset-objects). To retrieve an average of those particular months, you can request a quarterly frequency starting in May:

https://api.cal-adapt.org/api/series/tasmax_day_ens32avg_rcp85/cdd/?freq=QS-MAY&thresh=65&g=Point(-121.4688+38.59)

You can fine tune the number of months as well, so `2MS` would give you a bimonthly frequency.

clhenrick: The issue is that the current (v2) app allows users to select arbitrary months, so a user could select January and then June (which wouldn't make any sense). We would need to redesign that part of the UI to restrict it to quarterly periods or something more restrictive to align with the Pandas offsets, which I'm not sure our users would object to or not.

How does the API aggregate CCD & HDD? Is it summing them based on `freq`?

bkg: Ah, okay you could do any number of sequential months with the API, not limited to quarters, but you couldn't group non-sequential months with the resampler.

The default aggregation is `?stat=mean` if not specified, but more summary stats are available like min, max. The [docs](https://berkeley-gif.github.io/caladapt-docs/series.html#time-series) have some more detail.

clhenrick: I think it makes more sense to limit it to sequential months, but we would need to rethink the UI for the frequency selection which I'm not sure we have time for right now. Definitely worth brainstorming on. 

For the immediate time being if we continue with our current approach for monthly frequency by passing`freq=M`, and then filter and aggregate clientside for only the months the user has selected, would it be correct to also pass `stat=sum` and then average the results when aggregating client side?

clhenrick: Context: I'm trying to avoid computing an average of averages ([here's a write up on why that's a bad practice](https://convalytics.com/statistics/taking-an-average-of-averages-why-its-wrong/)). Because months differ in the number of days they have, we probably shouldn't be averaging the average number of CDD/HDD for each month by the number of months the user has selected. We should instead be averaging by the total number of CDD/HDD by the total number of days for all selected months.

mukhtyar: The `stat` param is related to spatial aggregation across grid cells and not temporal aggregation. So in case of temperature related indicators, if you want temperature for a polygon, e.g county, you get a temperature value for the county by taking the `mean` of temperatures across all intersecting grid cells. In case of precipitation amount for a county, it makes more sense to `sum` the amount of rain in each grid cell. On the other hand if you want rate of precipitation, then you might use `mean`. 

In the CDD/HDD calculations, for a grid cell, the values returned for each month/year (indicated by the `freq` param) is the total number of CDD/HDD for that month/year. This also applies to values returned for a county. So I think averaging the totals over `n` months in the tool is ok. 

When running the calculations for a county in the API, I think the spatial aggregation happens first. So we have a new daily timeseries of tasmax and tasmin for the county, and the CDD/HDD calculation & any temporal aggregation happens next on this new timeseries. (not sure if this is the sequence of events or if CDD/HDD values are calculated for each grid cell and then aggregated over the county; @bkg can confirm)

bkg: Sorry, I misspoke about the `stat` param for the CDD and HDD endpoints which differs from events. For CDD/HDD, `stat` controls spatial, not temporal, aggregation. You'll want to keep using mean for spatial aggregation, which is the default, so just passing `freq=M` should be enough.

clhenrick: Okay, got it. Thanks for clarifying @mukhtyar & @bkg

mukhtyar: > 
> 
> If you'd like to add ensembles for CDD and HDD, you can compute them dynamically from the various tasmax 32 GCM ensembles with varying thresholds and frequencies like:
> 
> https://api.cal-adapt.org/api/series/tasmax_day_ens32avg_rcp85/cdd/?freq=A&thresh=65&g=Point(-121.4688+38.59)

@bkg weren't there issues with calculating an envelope from the daily ensemble max and min `tasmax_day_ens32max_rcp85` and `tasmax_day_ens32min_rcp85`? It returned the most extreme values? Or was that only an issue for the Extreme Heat Day calculations? 

To show the ranges for Extreme Heat Days in the Location Climate Change Snapshot tool we ended up creating new timeseries `exheat_30y_ens32max_rcp85` and `exheat_30y_ens32min_rcp85` (using a fixed threshold)

bkg: That's correct. Generally it's an issue for the more complicated climate indices such as extreme heat, KBDI, but not the more straightforward ones like CDD and HDD which only rely on number of degrees above/below a static threshold instead of a percentile like extreme heat. 

If it's problematic, then it will be evident from the ensemble min/max as you mentioned, but I think it will be okay for these indices. Plus, it allows using a dynamic threshold which the pre-generated grids do not.

mukhtyar: Got it, thanks! We will try it out.

clhenrick:
@mukhtyar I think this is close. I'm not sure though if the values in the line chart are correct though, they're upwards in the thousands whereas in the v2 site they're only in the 100-200 range. 

I also haven't looked too deeply into handling the "no data" response, but I think in one of the utilities fetch helper functions it throws an error when no data is returned by the API. IMO that isn't technically an error, so maybe we shouldn't treat it as one and instead treat it as an empty state in the dashboard.

src/routes/tools/degree-days/_constants.js
```diff
@@ -0,0 +1,12 @@
+export const TOOL_SLUG = "degree-days";
```
mukhtyar, line 1: Like this idea of having a list of constants for the tool also

src/routes/tools/degree-days/_store.js
```diff
@@ -0,0 +1,61 @@
...
+  DEFAULT_SELECTED_MONTHS,
+} from "./_constants";
+
+// List of climvars used in Degree Days Tool
+export const climvarList = climvars
+  .filter(({ id }) => id === CLIMATE_VARIABLE)
+  .map((d) => ({ ...d, title: `Degree Days ${d.label}` }));
+
+// Q: should this even be a store if it's only ever a single value and doesn't change?
+export const climvarStore = (() => {
```
mukhtyar, line 18: I don't think you need a climvar store for this tool.

src/routes/tools/degree-days/_data.js
```diff
@@ -0,0 +1,239 @@
...
+  method = "GET",
+  indicatorId,
+  monthIds,
+  isEnsemble,
+}) => {
+  const url = `${apiEndpoint}/series/${slug}/${urlPathAfterSlug(
+    indicatorId,
+    isEnsemble
+  )}/`;
+  const [response, error] = await handleXHR(fetchData(url, params, method));
```
mukhtyar, line 100: On thinking some more, we probably don't need to use `handleXHR` here. I wrote that function to track errors when users upload their own boundary for aggregation (because there are a whole bunch of different error messages returned by the API). But now all that checking first happens in the `tools/partials/ChangeLocation`, so by the time the geometry reaches `fetchEvents` it's geometry that's good to go. Instead we can just check for 404 errors (incorrect slugs) or empty arrays. If the array is empty we can show the No Data state.

src/routes/tools/degree-days/_data.js
```diff
@@ -0,0 +1,239 @@
...
+
+/**
+ * Fetches data for a single observed/model/ensemble series
+ * Each series can have 1 or more slugs
+ * @param {object} series - a distinct timeseries obj
+ * @param {object} params - props for for geometry, stat, units, etc.
+ * @param {string} method - default is GET, POST for uploaded boundaries
+ * @return {array}
+ */
+const fetchSeries = async ({
```
mukhtyar, line 120: This function might be a good spot to backfill any missing data with 0. The year ranges for the `livneh (1950-2005)` and `models (2006-2099)` are standard and can be added to the list of constants

src/routes/tools/degree-days/_data.js
```diff
@@ -0,0 +1,239 @@
...
+    const values = merge(responses);
+    if (!values.length) {
+      throw new Error(`${series.id}: No Data`);
+    }
+    // For livneh, remove data values after 2006
+    // because there are QA/QC issues with the data
+    if (series.id === "livneh") {
+      return {
+        ...series,
+        values: values.filter((d) => d.date.getFullYear() <= 2006),
```
mukhtyar: Sorry I made a mistake here, this should be `< 2006`

src/routes/tools/degree-days/_constants.js
clhenrick: Thanks, I think of it sort of like a local config file for the tool for variables that are unique in value but common to all tools.

src/routes/tools/degree-days/_store.js
clhenrick: It's definitely overkill and could remove it, but it gets destructured in ExploreData and is referenced various places there. It might be worth keeping just for maintaining consistency between tools.

src/routes/tools/degree-days/_data.js
clhenrick: So swap it and use `window.fetch`? 

For the empty data state we will need to rethink the logic in ExploreData. Currently it will show the loading spinner if `$dataStore` is an empty array. We might just want to hide the loading after `$dataStore` changes to no longer be `null`. Or better yet, have a separate `isLoadingData` Svelte Store that is a boolean which is toggled to `true` before a request is made and then toggled to `false` after a request has finished or errored. That way we can show a loading spinner whenever data is fetched, but that might not be necessary if data has already been cached by the browser.

src/routes/tools/degree-days/_data.js
clhenrick: Ah right, I had forgotten about that part! Will make sure to backfill for any missing dates. 

We may also want to consider a `0` value and `no data` value separately and not synonymous, as technically zero is a value while missing data is just missing data.

src/routes/tools/degree-days/_ExploreData.svelte
```diff
@@ -141,6 +139,12 @@
 
   $: formatFn = format(`.${$climvar.decimals}f`);
 
+  $: indicatorTitle = $indicator.title.replace("Degree Days ", "");
```
mukhtyar: The chart title needs to indicate Heating Degree Days or Cooling Degree Days.

src/routes/tools/degree-days/_data.js
mukhtyar: Adding a separate isLoadingData store is a good idea.

src/routes/tools/degree-days/_ExploreData.svelte
clhenrick: This is for the MadLib next to the locator map, the chart title is created in the `yAxis.label` prop.

src/routes/tools/degree-days/_data.js
clhenrick: Cool. The one thing to be cautious of from a UX perspective is to not flash a loading spinner very quickly after the settings change. It can be triggering and also make the data update seem like it's taking longer than it is.

src/routes/tools/degree-days/_ExploreData.svelte
mukhtyar: MadLib 😄 . That's a perfect description. The Heating/Cooling needs to be in the madlib also.

clhenrick: Yeah, that's how I think of these types of dynamic sentences :)

It's there, I'm just removing an extra "Degree Days" from the title:

Screen Shot 2021-09-17 at 4 11 47 PM

Screen Shot 2021-09-17 at 4 11 56 PM

clhenrick: image

PR #31, opened 2021-09-17
Update of Glossary Terms with Finalized Language

elehmer:
Here are the updated MD files for the Glossary Section with the finalized language. Let me know if you want me to update the MD styling. Reference links needed to be in HTML instead of MD because they are inside a <P> tag.


clhenrick (approved):
Looks good to me @elehmer thanks for this!

PR #33, opened 2021-09-17
Updated FAQs Markdown Files with Tags and Categories Added

elehmer:
Here are the FAQ markdown files from the finalized document. Let me know if more styling is needed. 

I added tags and a new property category. Please review the tags and adjust to what is needed for site. The categories are suggestions that you can use to make the final version of. One thing I realized is that we don't have an ordering within a category so if you want me to add a rank ordering property let me know. Happy to do any editing of this stuff to unload your burden.


content/faqs/how-can-i-download-boundary-layers.md
```diff
@@ -0,0 +1,49 @@
+---
+title: How can I download boundary layers?
+tags: data
+category: getting-data
+---
+
+The following boundary layers can be downloaded from the Cal-Adapt API in GeoJSON format:
+
+
+[California Climate Change Fourth Assessment Regions](https://api.cal-adapt.org/api/ccc4aregions/)\
```
clhenrick, line 10: @elehmer I think listing the URL after the link is redundant. Could we just turn these into an unordered list and drop the extra URL text?

content/faqs/how-can-i-get-help.md
```diff
@@ -1,10 +1,11 @@
...
 ---
 
-All Cal-Adapt climate tools include **Learn More** links that provide explanations and more information about tool parameters and terminology. 
+All of Cal-Adapt’s climate tools include “Learn More” links that provide explanations and more information about tool parameters and terminology.
 
 You can also explore our [Getting Started](/help/get-started/) guide or [Glossary](/help/glossary/) to find an answer to your question.
 
-For general help, you can email support@cal-adapt.org. Cal-Adapt staff frequently monitor this email address and are ready to support users in accessing California climate data.
\ No newline at end of file
+For general help, you can email <support@cal-adapt.org>. Cal-Adapt staff frequently monitor this email address and are ready to support users in accessing California climate data.
```
clhenrick, line 11: Nice, I didn't know you can make a `mail:to` link by doing `<email@example.com>` like this with markdown

content/faqs/can-i-get-data-for-my-aoi.md
```diff
@@ -1,7 +1,7 @@
 ---
-title: Can I get data for my project area?
+title: Can I get data for my specific project area?
 tags: data
+category: getting-data
```
clhenrick, line 4: thanks for consistently naming the categories using this shorthand. We can map it to the more human readable strings in the app.

clhenrick (changes requested):
@elehmer thanks for this! One small request to clean up the list of data downloads in "How can I download boundary layers?" (see my note below)

content/faqs/how-can-i-download-boundary-layers.md
elehmer: I think the intent here is that URL be visible so that you can easily copy it into code, but maybe just better to make them jump one extra step. Thoughts?

content/faqs/how-can-i-get-help.md
elehmer: Yeah I was surprised it handled in like that. Might be a browser thing so we should probably test this across browsers.

content/faqs/can-i-get-data-for-my-aoi.md
elehmer: No problem.

content/faqs/how-can-i-download-boundary-layers.md
clhenrick: Could be worth verifying, but if that's the case it could be helpful to make the plain text URLs code blocks, like:

[California Climate Change Fourth Assessment Regions](https://api.cal-adapt.org/api/ccc4aregions/)
```
https://api.cal-adapt.org/api/ccc4aregions/
```

clhenrick: But IMO it's still redundant as you can easily right click on the link and "copy URL"

elehmer: True. Ok, I will edit these out...

content/faqs/how-can-i-get-help.md
clhenrick: It compiles to `<a href="mailto:support@cal-adapt.org">support@cal-adapt.org</a>` so I think we'll be fine on different browsers.

content/faqs/how-can-i-download-boundary-layers.md
clhenrick: I think we should call this good for now and maybe later come back to it.

PR #34, opened 2021-09-20
Add extreme heat tool

mukhtyar:


src/routes/tools/extreme-heat/_ExploreData.svelte
```diff
@@ -0,0 +1,460 @@
...
+  <div slot="settings" class="settings">
+    <div class="block">
+      <RadioBtnGroup
+        selected="{$climvarStore}"
+        items="{climvarList}"
+        title="Select Indicator"
+        on:change="{changeClimvar}"
+      />
+      <SelectClimvar
+        title=""
```
clhenrick: Perhaps we should add a title for this dropdown? It's not clear to me what it's for. Adding a title would help make it more noticeable in the Settings UI as well.

src/routes/tools/extreme-heat/_ExploreData.svelte
```diff
@@ -0,0 +1,460 @@
...
+  }
+</style>
+
+{#if isLoading}
+  <Loading />
+{/if}
+
+<Dashboard>
+  <div slot="map">
+    <StaticMap location="{$location}" width="{500}" height="{500}" />
```
clhenrick: I'm noticing that when I select the whole state as the aggregation boundary the map fails to display. Is this an issue for all tools?

image

clhenrick: Also after selecting the state, when I change / toggle the indicator from extreme heat to warm nights it doesn't seem to update the chart.

clhenrick: And if I change the scenario I get an error that breaks the dashboard:

image

src/routes/tools/extreme-heat/_ExploreData.svelte
```diff
@@ -0,0 +1,460 @@
...
+        selectedId="{$scenarioStore}"
+        items="{DEFAULT_SCENARIOS}"
+        on:change="{changeScenario}"
+      />
+      <LearnMoreButton
+        on:click="{() => loadLearnMore({ slugs: ['emissions-scenario'] })}"
+      />
+    </div>
+    <div class="block">
+      <SelectThreshold
```
clhenrick: Noticing a few things with this UI that makes it feel a bit awkward / confusing:

1. Having a single radio button isn't really a good use of radio buttons. Maybe we can list two or three options instead to make it clear that choosing additional thresholds is an affordance?
2. Currently the way in which additional thresholds are added is a little too responsive. Maybe it would be better to have a "Add Threshold" button next to or below the text input that would need to be clicked on in order to add it to the list?
3. If I delete a custom threshold that is also the currently selected threshold, the chart and stats retains the same state. Perhaps in this case the selected threshold should revert to the default 98th percentile threshold?
4. In this case I think the help text would actually be better placed below the "Select Threshold" title, as it refers not just to the input but also to the radio buttons.

clhenrick: Also the custom threshold doesn't seem to be captured in the Share / Bookmark.

src/routes/tools/extreme-heat/_ExploreData.svelte
clhenrick: Noticing that this value is not captured in the Share / Bookmark URL.

src/components/tools/Charts/Heatmap/Chart.svelte
```diff
@@ -30,32 +26,38 @@
     tickFormat: timeFormat("%Y"),
     units: "",
   };
+  export let colors = ["#fed976", "#fd8d3c", "#e31a1c", "#800026"];
```
clhenrick: Should these colors be pulled from a constant file or CSS vars somewhere? Not a big deal but helpful to track where we're using various colors to help maintain consistency.

src/components/tools/Charts/Heatmap/Chart.svelte
```diff
@@ -30,32 +26,38 @@
...
+    initialSeriesSelection = data.find(({ values }) => {
+      const seriesXMax = max(values, (d) => d.date);
+      return seriesXMax.getFullYear() >= xmax.getFullYear();
+    });
+
+    // Update visibility of data series
+    chartData = data.map((series) => {
+      return {
+        ...series,
+        visible: series.id === initialSeriesSelection.id ? true : false,
```
clhenrick: you can simplify this by just doing `visible: series.id === initialSeriesSelection.id` and dropping the `? true : false`

src/components/tools/Charts/Heatmap/Chart.svelte
```diff
@@ -89,38 +89,22 @@
...
-    series = data.map((d) => {
+    // Set Controls
+    const controlItems = data.map(({ id, label, color }) => {
       return {
-        key: d.key,
-        label: d.label,
+        id,
+        label,
+        color,
+        visible: id === initialSeriesSelection.id ? true : false,
```
clhenrick, line 98: ditto with simplifying the ternary here. 

Also a way to simplify returning an object with an arrow function is to wrap the object in paranthesis. That way you can drop the `return` and extra curly braces:

```js
const controlItems = data.map(({ id, label, color }) => ({
  id,
  label,
  color,
  visible: id === initialSeriesSelection.id,
})
```

Or using the spread operator (assuming you want to keep all the props):

```js
const controlItems = data.map((d) => ({ ...d, visible: id === initialSeriesSelection.id })
```

src/components/tools/Charts/Heatmap/Control.svelte
```diff
@@ -0,0 +1,69 @@
+<script>
+  import { getContext } from "svelte";
+
+  const controlItems = getContext("Control");
+
+  const toggleDisplay = (id) => {
+    return [...$controlItems].map((series) => {
+      return {
+        ...series,
+        visible: series.id === id ? true : false,
```
clhenrick, line 10: ditto here with simplifying the ternary

src/components/tools/Charts/Scatter/Chart.svelte
```diff
@@ -60,7 +60,7 @@
     legendItems.set(
       data.map((series) => {
```
clhenrick: is this just pulling certain props from `data`? If so you could just destructure since all the prop names are the same in returned object.

src/components/tools/Settings/SelectThreshold.svelte
```diff
@@ -67,27 +60,19 @@
...
-          style="padding-right:0.5rem;width:50%;"
-          type="number"
-          id="threshold-input"
-          step="1"
-          on:keydown="{addThreshold}"
-        />
-      </div>
-    </div>
+  <div style="margin-top: 0.5rem;">
+    <NumberInput
```
clhenrick: I think what might work better here is wrapping this in a `<form>` (or whatever the Carbon component equivalent is) and adding a submit button with the text "Add Threshold" (see my other comment in ExploreData on why).

src/routes/tools/extreme-heat/_constants.js
```diff
@@ -0,0 +1,70 @@
...
+
+export const CLIMATE_VARIABLES = ["tasmax", "tasmin"];
+export const CLIMATE_INDICATORS = [
+  {
+    id: "frequency",
+    label: "Frequency",
+    title: "Number of Extreme Heat Days per Year",
+    units: "days/yr",
+    decimals: 0,
+    chartComponent: LineAreaChart,
```
clhenrick, line 18: Interesting way of solving how to toggle the chart and stats panel types. If I had done this I might have separated the components from the climate indicator property list because they're differing data types but this makes it simpler when toggling between indicators.

src/routes/tools/extreme-heat/_data.js
```diff
@@ -1,37 +1,192 @@
...
+ * @return {number} thresh98p
+ */
+
+export async function get98pThreshold(climvar, params) {
+  try {
+    const { id } = climvar;
+    const url = `${apiEndpoint}/series/${id}_day_livneh/exheat/?${serialize(
+      params
     )}`;
     const data = await fetch(url);
```
clhenrick, line 191: You can simplify this by doing something like:

```js
return +(await (await fetch(url)).json())["98p"].toFixed(1);
```

It's a trade off between readability and the number of local variables, but it can be helpful for reducing code bloat and memory allocation sometimes by limiting the number of local variables when writing procedural code like this.

src/routes/tools/extreme-heat/_data.js
```diff
@@ -42,113 +197,76 @@ export async function get98pThreshold(climvar, queryParams) {
...
-      key: "observed",
-      type: "line",
-      values: values.filter((d) => d.date.getFullYear() < 2007),
-    };
-    return addSeriesInfo(observed);
-  } catch (error) {
-    throw new Error(`Observed: ${error.message}`);
+// Group daily data by year
+// Backfill data for missing years with empty array
+export const groupDataByYear = (series) => {
```
clhenrick, line 207: FYI I added a utility function in `~/helpers/utilities.js` for backfilling the years. But should we clarify with Brian that we should be doing this client side? For degree days when we asked him about it we found out there was a bug in the data returned by that particular API endpoint, we might want to double check here as well.

clhenrick (changes requested):
Noticed a bug in the Select Location UI that you might already be addressing. I do think we should modify the implementation of the Select Threshold setting UI though as it's a little too responsive in its current implementation (see comments below). Let me know if you'd like to chat about it or punt it to later though.

src/routes/tools/extreme-heat/_ExploreData.svelte
mukhtyar: Aggregating by CA state boundary is not supported for any tool that uses daily data (e.g. Extreme heat, CDD/HDD). The #38  pr should fix this by making this the default.

src/routes/tools/extreme-heat/_data.js
mukhtyar: I'm using the `events` endpoint for the Extreme heat tool which returns the raw data (in this case - temperature) for all dates. With the `thresh` param it will return data values only above the threshold. I calculate all the indicators locally, e.g. count/max duration, etc. So it behaves a little differently than the `exheat`/`cdd`/`hdd` endpoints. I think in this case it should not be backfilled with 0 from the API.

clhenrick: For the instances of `new Date(year, 0, 1)` it might be good to do `new Date(Date.UTC(year, 0, 1))` instead to avoid timezone issues.

src/components/tools/Settings/SelectThreshold.svelte
mukhtyar: Fixed via https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/656c47668e2e408e9d11d23505e2300b0d60c537

src/routes/tools/extreme-heat/_ExploreData.svelte
mukhtyar: I've removed the ability to remove custom thresholds as it's not in the current tool via https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/907dd706e07ebe99fb64cff539501d939c0f18c5. This will simplify a lot of the user interaction. Also added 2 commonly used alternate values so the list folks can pick from is not just 1 radio button via https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/57644c2547ef9e25ed64b7a28acbacfa36ab3691

src/routes/tools/extreme-heat/_ExploreData.svelte
mukhtyar: Added thresh to bookmark in https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/4bc32af5a9f1d0a6de98bb9a4d3f16cb3a1cfc5a. Added title Select View to dropdown in https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/fdcb3248d4d20f46b58d17fc856f5fe97d81d223

src/components/tools/Charts/Heatmap/Chart.svelte
mukhtyar: Good idea. Added constant `HEATMAP_COLOR_SCALE` in https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/d0ad393a82a20f46a72cc00cf8d6add83d278fd9

src/components/tools/Charts/Heatmap/Chart.svelte
mukhtyar: Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/914278c544232c4ccea2c9440a6e16053281f7e6

src/components/tools/Charts/Scatter/Chart.svelte
mukhtyar: Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/b7fdecf54b22a34d0e8e224c46d8321abb24feef

src/routes/tools/extreme-heat/_data.js
mukhtyar: I find this hard to read :-) so I made a minor simplification in https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/3f2f836c6c2309c24915d30add4fc7b3960411ca

src/routes/tools/extreme-heat/_data.js
mukhtyar: Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/34/commits/f9f62a18b2fcf13ce3dcaabcacdd0c8b021cbba9

src/routes/tools/extreme-heat/_ExploreData.svelte
clhenrick: @mukhtyar when I try to load the bookmarked URL with a custom threshold I get an error and the dashboard doesn't load: `init error TypeError: $store is undefined`

mukhtyar: I've removed `thresh` from the bookmark. It's tricky to sync a custom `thresh` when the tool first loads, since the `thresh` controls the data returned from the api and is reset each time the `location` or `climvar` changes. Since it's not a part of the functionality offered in the current tool, it's probably not a good use of time to figure out all the edge cases.

PR #37, opened 2021-09-21
Fix select model settings

mukhtyar:
Fixes CSS vars and updates labels to align with the carbon MultiSelect component

### DIFF
--- a/src/scss/site/_dashboard.scss
+++ b/src/scss/site/_dashboard.scss
@@ -5,62 +5,62 @@
 .dashboard {
   input[name*="ACCESS1-0"] + .bx--checkbox-label::before,
   input.bx--checkbox[name*="ACCESS1-0"]:checked + .bx--checkbox-label::before {
-    background-color: --gcm-ACCESS1-0;
-    border-color: --gcm-ACCESS1-0;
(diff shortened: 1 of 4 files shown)


src/scss/site/_dashboard.scss
```diff
@@ -5,62 +5,62 @@
 .dashboard {
   input[name*="ACCESS1-0"] + .bx--checkbox-label::before,
   input.bx--checkbox[name*="ACCESS1-0"]:checked + .bx--checkbox-label::before {
-    background-color: --gcm-ACCESS1-0;
-    border-color: --gcm-ACCESS1-0;
+    background-color: var(--gcm-ACCESS1-0);
```
clhenrick: Not a big deal to use the CSS variables but couldn't you use the Sass equivalent of these variables here?

clhenrick (approved):
Just a question about the CSS vars, otherwise looks good

src/scss/site/_dashboard.scss
mukhtyar: Fixed via https://github.com/berkeley-gif/caladapt-website-2021/pull/37/commits/32af8060f52a331204ef8d1cee5d519a0aea1f22

PR #39, opened 2021-09-21
Add v2 links

mukhtyar:
Add links to v2 cal-adapt for some Tools and all Data Download instances


clhenrick (approved):
Thanks Shruti, the Local Climate Change Snapshot tool still gives a 404, do we want to fix it in this PR?

clhenrick:
@mukhtyar actually there's one other place where we should fix the data download link, in `content/faqs/how-can-i-download-data-from-caladapt.md`

mukhtyar:
> Thanks Shruti, the Local Climate Change Snapshot tool still gives a 404, do we want to fix it in this PR?

Hmm, I don't see it? The link should be `https://v2.cal-adapt.org/tools/local-climate-change-snapshot/`

clhenrick:
@mukhtyar it must not have finished rebuilding when I checked locally, I see that it points to the v2 domain now.

PR #46, opened 2021-09-22
Add under construction placeholders

mukhtyar:


clhenrick (approved):
Thanks @mukhtyar for doing this. I'm wondering if we could set some temporary CSS to at least make the page the height of the browser viewport so we don't end up with the footer in the middle of the page? Perhaps we could move the notification to below the header in the body (like what you did with the data download tool) but not a big deal if not.

One thing the tools has me thinking about because there is so much common markup between them is if there's a way we could use a `[slug]` dynamic route for them? Or would this make customizing each tool more difficult? Maybe one way could be having all of the common stuff in a `src/routes/tools/[slug].svelte` and then having separate `_tool-name/` directories that contain the custom stuff? We could always revisit this later after migrating them too, just interested on your thoughts.

mukhtyar:
Yeah, there is still a lot of common markup & functionality in the `preload` script  in the tools. A good time to revisit this might be when we migrate to `sveltekit`.

PR #50, opened 2021-09-22
Added order tag to FAQs so that categories can be sorted

elehmer:
I have added an order tag 1..n for each category so that FAQs can be ordered as they appear in the FAQ word document. The category functionality needs to be added to the FAQs still.

### DIFF
--- a/content/faqs/can-i-get-data-for-my-aoi.md
+++ b/content/faqs/can-i-get-data-for-my-aoi.md
@@ -2,6 +2,7 @@
 title: Can I get data for my specific project area?
 tags: data
 category: getting-data
+order: 3
 ---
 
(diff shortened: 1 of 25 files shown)


mukhtyar:
What do you think about storing the longer version of the text in `category` e.g. "Understanding, using, and interpreting data"? That way we don't have to reference it somewhere else. Is the current ordering intentional? I'm not sure we need the `order` prop, it's also hard to maintain. The FAQs can just be sorted alphabetically (as a whole or for each category) when the markup to json transformation takes place.

clhenrick:
@elehmer was this a request from Nancy? I don't recall discussing adding an `order` property to the FAQs.

clhenrick:
Closing for reasons discussed on chat.

PR #64, opened 2021-09-27
Add beta site alert

mukhtyar:


mukhtyar:
[Create an alert message for Beta site](https://trello.com/c/hNvHKarg/97-create-an-alert-message-for-beta-site)

src/routes/_layout.svelte
```diff
@@ -13,6 +14,8 @@
   <a href="#main-content">Skip to Main Content</a>
 </div>
 
+<SiteAlert />
```
clhenrick: @mukhtyar can we only append this when "beta" is in the subdomain? Or via a flag via `process.env.BETA`? Might be helpful to not see this while developing locally and on the dev site.

src/partials/SiteAlert.svelte
```diff
@@ -0,0 +1,30 @@
...
+  <InlineNotification
+    lowContrast
+    hideCloseButton
+    kind="info"
+    title="This is a preview of the new Cal-Adapt website."
+    subtitle="Return to cal-adapt.org to access all climate tools and data."
+  >
+    <div slot="actions">
+      <NotificationActionButton>
+        <Link href="https://cal-adapt.org/">cal-adapt.org</Link>
```
clhenrick: Maybe we can say "Return to Cal-Adapt" or "Go to Cal Adapt" here for the link text? I think ideally it would help to make the subtitle prop in the `InlineNotification` component a link but that doesn't seem like an option with this component.

clhenrick (approved):
IMO it would be helpful to only show this on `beta.cal-adapt.org` so that when we're developing we don't get bothered by it or have to disable it.

src/partials/SiteAlert.svelte
mukhtyar: The `subtitle` prop seems to accept only text. That would have been ideal. I will update the link text.

src/routes/_layout.svelte
clhenrick: @mukhtyar if you merge recent changes from `main` into this branch you should be able to  check for `process.env.showBetaBanner`

mukhtyar: Fixed via https://github.com/berkeley-gif/caladapt-website-2021/pull/64/commits/79841d5f0932e6a763e25560e9495a3820b8fe5a

src/partials/SiteAlert.svelte
mukhtyar: Fxed via https://github.com/berkeley-gif/caladapt-website-2021/pull/64/commits/79841d5f0932e6a763e25560e9495a3820b8fe5a

PR #66, opened 2021-09-28
Add github action for nodejs & sapper export

clhenrick:
Set up a Github Action that will run on all future PRs. If this isn't helpful we can always adjust or remove it later.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -14,17 +14,12 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [12.x, 14.x, 16.x]
-        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
-
     steps:
     - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


clhenrick:
Because the `sapper export` command takes a while we may only want to test this on one version of NodeJS, but we might want to agree on what NodeJS version we're using.

clhenrick:
@mukhtyar @elehmer I'm curious as to what version of NodeJS you both are using? I'm on a LTS `14.17.x`

mukhtyar:
`14.17.0` and `14.16.0`. I can update to `14.17.x` if needed.

elehmer:
`12.22.6`...I guess it should be more up-to-date than that? I didn't know what you guys were running so I just got it up to the minimum needed to build. Let me know what versions you are on and I'll update.

clhenrick:
Thanks @mukhtyar @elehmer maybe we can stick with `14.17.x` which is a LTS version of NodeJS. It occasionally helps with bugs if we're all on the same version.

elehmer:
Ubuntu 20.04 gravitated to 14.18.x as its LTS. I'm not going to battle to get to 14.17.x as it shouldn't cause any issues I don't think.

PR #67, opened 2021-09-28
Fix scroll pos when navingating from home to signup pages

clhenrick:


clhenrick:
[Bug: Window scroll is off when going to signup from homepage](https://trello.com/c/E9D8udHJ/98-bug-window-scroll-is-off-when-going-to-signup-from-homepage)

mukhtyar (approved):

PR #73, opened 2021-09-29
Replace with smaller image

mukhtyar:


clhenrick (approved):
@mukhtyar 300kb and less is way more reasonable 👍 

One other thing I might suggest, and please don't worry about doing it here if you don't feel like it, is for higher resolution screens it's typically good to size the image 2x the actual width it will be rendered at. So for example in the content area of the blog if the images will only ever get as big as 900px wide then I might size them at 1800px wide, and then adjust the image quality using the settings for JPG to keep the size reasonably low. However sometimes this is still too large for mobile, which is why it's good to have at least one other image size for smaller devices (e.g. if the max width were 500px wide then have a separate 1000px wide image that is displayed at mobile breakpoints). 

But I say this with a grain of salt as the way we're "supposed to" optimize images for the web keeps changing and I'm not 100% up to date on what the best methods are. At the very least I think it's good to keep the image sizes as low as possible without making them look awful though.

mukhtyar:
Thanks @clhenrick. I was thinking of creating a couple of blog templates. This information on image sizes and resolution will be very helpful as part of the templates for future reference.

PR #74, opened 2021-09-30
Use state boundary from api

mukhtyar:
This PR updates the tools and Location & Boundary related components to use the state boundary vector layer from the api. There is additional cleanup to do in the `src/helpers/geocode,js` . Proposing to update it after the Extreme Heat and Snowpack tools are migrated so sapper does not complain about missing functions.


src/components/tools/Location/StaticMap.svelte
```diff
@@ -34,41 +31,55 @@
...
-        const simplifiedGeojson = simplify(geojson, {
-          tolerance: 0.01,
-          highQuality: true,
+      // Simplify geometry for more complex shapes
+      simplifiedGeojson = simplify(geojson, {
+        tolerance: 0.005,
+        highQuality: true,
+      });
+      // Simplify further for even more complex shapes (e.g. state boundaries)
+      if (JSON.stringify(simplifiedGeojson).length > 8000) {
```
clhenrick: This seems like it might be a good use for recursion? Like recursively simplify the geometry until it's at the desired tolerance. That would catch any cases where this `if` clause passes but the geometry is still too big.

src/components/tools/Location/StaticMap.svelte
```diff
@@ -34,41 +31,55 @@
...
       }
-      bounds = "auto";
+      overlay = encodeURIComponent(JSON.stringify(simplifiedGeojson));
+    }
 
-      src = `https://api.mapbox.com/styles/v1/${style}/static/geojson(${overlay})/${bounds}/${width}x${height}?padding=${padding}&access_token=${mapboxgl.accessToken}`;
+    return `https://api.mapbox.com/styles/v1/${style}/static/geojson(${overlay})/${bounds}/${width}x${height}?padding=${padding}&access_token=${mapboxgl.accessToken}`;
+  }
+
+  $: if (location) {
```
clhenrick: Just a suggestion, feel free to take it or leave it: I would simplify this by taking the handling of creating the point url and wrapping it in a function like `createPointImgSrc()` and then turn this into a `if else`

```js
$: if (location && location.geometry) handleLocation(location);

function handleLocation(feature) {
  if (feature.geometry.type === "Point") {
    createPointImgSrc(location);
  } else {
    createPolygonImgSrc(location);
  }
}
```

It would help improve readability and make it easier to follow the logic IMO by avoiding the nesting of `if`'s and using a function in one block but not the other.

You could also have a separate function that creates the URL string so that you don't have to duplicate the code for it for each geometry type.

clhenrick (approved):
@mukhtyar just some minor suggestions, feel free to take them or leave them.

src/components/tools/Location/StaticMap.svelte
mukhtyar: Good idea, fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/74/commits/182e8c7c41953a742e87da1ff4b8c54e5d2d9be6

src/components/tools/Location/StaticMap.svelte
mukhtyar: Recursion makes my head hurt :-) but I tried it out in  https://github.com/berkeley-gif/caladapt-website-2021/pull/74/commits/60ef734801f76afb5a70bc89563e64cd045dbee6. Let me know what you think.

mukhtyar:
@clhenrick I'm done adding the `ImageLoader` carbon component. Can you take another quick look at it when you have time?

clhenrick:
@mukhtyar sure thing, will take a look right now.

src/components/tools/Location/StaticMap.svelte
```diff
@@ -34,84 +35,84 @@
...
+    // The Mapbox Static Images API only accepts requests that are 8,192 or fewer characters long.
+    // Truncate coordinate precision
+    const truncatedGeojson = truncate(geojson, { precision: 4 });
+    // Simplify geometry
+    const overlay = createOverlay(truncatedGeojson);
+    return createSrcUrl({ overlay, bounds, params });
+  }
+
+  function handleLocation(feature) {
+    alt = `map of ${feature.title}`;
```
clhenrick: I personally wouldn't set the value of `alt` here and might instead make it a reactive variable like 

```js
$: alt = location ? `map of ${location.title}` : ""
```

An empty string (`""`) is a valid `alt` text attribute and indicates to screen readers to ignore the image. 

I think you're also using both a local variable and global variable for the same thing within this function. It would make it easier to follow to just use the local variable `feature` IMO.

src/components/tools/Location/StaticMap.svelte
```diff
@@ -11,15 +14,13 @@
   export let style = "mapbox/streets-v11";
   export let padding = 50;
   export let zoom = 8;
-  export let alt = "location map";
 
+  const { accessToken } = mapboxgl;
   let img;
```
clhenrick: Because you're now using the  `ImageLoader` component I think you can get rid of these unused variables (img, loading, error) as well as remove the `load()` function and the `if` statement that calls it below.

clhenrick (approved):
@mukhtyar just some suggestions for cleaning up the old code now that you're using the image loader component.

src/components/tools/Location/StaticMap.svelte
mukhtyar: Fixed via https://github.com/berkeley-gif/caladapt-website-2021/pull/74/commits/c9e86d7a69f548af26546dedbdca05b9676b944f

PR #75, opened 2021-10-05
Add cross-env to dev script

mukhtyar:
Adding `cross-env` library so the `npm run dev` runs on my Windows machine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2321,15 +2321,6 @@
         "sha.js": "^2.4.8"
       }
     },
-    "cross-env": {
-      "version": "7.0.3",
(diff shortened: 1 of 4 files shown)


clhenrick (approved):
@mukhtyar probably not a bad idea to have this for compatibility across windows and linux/mac systems. 

Though instead of setting the `NODE_ENV="development"` in the npm script, we could also change all instances of `process.env.NODE_ENV === "development"` to `process.env.NODE_ENV !== "production"` as they essentially do the same thing (I think the only other value that is often checked for is "test" but we don't have any tests to run, at least not yet!) I think this only occurs in `src/routes/_error.svelte` and `src/routes/index.svelte` but worth double checking.

Note that in the `deploy.mjs` script we don't need to rely on `cross-env` because `zx` handles it for us which is nice.

PR #78, opened 2021-10-07
Create new explore scss partial

mukhtyar:


src/scss/site/_explore-data.scss
```diff
@@ -0,0 +1,64 @@
+#explore {
```
clhenrick: Could we namespace this under `#explore-data`? That way it’s more clear that it matches the name of the ExploreData component

clhenrick (approved):
@mukhtyar just a suggestion to change the namespace / id to explore-data from explore for clarity.

src/scss/site/_explore-data.scss
mukhtyar: Good idea. Updated namespace in https://github.com/berkeley-gif/caladapt-website-2021/pull/78/commits/265c1eadf53146cec76622c7e86a6bee4c99f438. Also deleted a tool related scss partial we not longer use.

PR #84, opened 2021-10-12
Address qa3 degreedays tool

mukhtyar:


clhenrick (commented):
@mukhtyar as I mentioned today, running into a similar issue with the locationStore when updating the chart's title in the snowpack tool. The way I fixed it (fingers crossed) that you might want to try here is updating the reactive variable in the [`afterUpdate` lifecycle method](https://svelte.dev/docs#afterUpdate). 

For example:

```js

let src;

afterUpdate(() => {
  if (location) {
    src = location.geometry ? handleLocation(location) : "";
  }
})
```

I mention it because I'm still experiencing the same issue locally with the StaticMap not updating when running this code.

clhenrick:
Actually @mukhtyar I just tried this and using `afterUpdate` doesn't seem to solve it. What does seem to solve it is replacing the `ImageLoader` component with a regular `` tag 🤷

I also used reactive variables for `src` and `alt` rather than reactive statements via `$:`

mukhtyar:
@clhenrick Thanks for testing out various strategies! I switched back to using the `img` tag and adding a reactive statement for `loadImage` which is called each time `src` changes. Possible that svelte was grouping the image update in a separate batch of changes and not right after the `src` value change. Hoping this fix https://github.com/berkeley-gif/caladapt-website-2021/pull/84/commits/34bd4a802e40992ba52c3f2a63a94324c8e83815 takes care of that.

src/components/tools/Location/StaticMap.svelte
```diff
@@ -16,6 +17,23 @@
...
   const { accessToken } = mapboxgl;
+  // Set to `true` when `loaded` is `true` and `error` is false
+  let loading = false;
+  // Set to `true` when the image is loaded
+  let loaded = false;
+  // Set to `true` if an error occurs when loading the image
+  let error = false;
+  let image;
+
+  const loadImage = (url) => {
```
clhenrick: instead of handling the image loading work in that function, you can just declare the image as a top level const variable  like:

```js
  const image = new Image();
  image.onload = () => {
    loading = false;
    loaded = true;
  }
  image.onerror = () => {
    loading = false;
    error = true;
  }
```

src/components/tools/Location/StaticMap.svelte
```diff
@@ -75,6 +93,7 @@
 
   $: src = location && location.geometry ? handleLocation(location) : "";
   $: alt = location ? `map of ${location.title}` : "";
+  $: if (src) loadImage(src);
```
clhenrick: and instead of using the reactive statement, use the `afterUpdate` lifecycle like:

```js
  afterUpdate(() => {
    if (src) {
      image.src = src;
    }
  });
```

But you might be able to just do:

```js
$: if (src) image.src = src
```

src/components/tools/Location/StaticMap.svelte
```diff
@@ -97,12 +116,19 @@
...
 <button class="static-map" on:click>
-  <ImageLoader src="{src}" alt="{alt}" fadeIn>
-    <div slot="loading">
-      <InlineLoading description="Loading location map..." />
-    </div>
-    <div slot="error">
-      <span class="error-text">An error occurred. Unable to load map.</span>
-    </div>
-  </ImageLoader>
+  {#if loading}
```
clhenrick, line 119: I think this would make more sense / be easier to follow if you used `if / else` here:

```svelte
<button class="static-map" on:click>
  {#if loading}
    <InlineLoading description="Loading location map..." />
  {:else if loaded}
    <img
      {...$$restProps}
      style="width: 100%;{$$restProps.style}"
      src="{src}"
      alt="{alt}"
      transition:fade
    />
  {:else if error}
    <div class="error-text">An error occurred. Unable to load map.</div>
  {:else}
    <p>Something went wrong.</p>
  {/if}
</button>
```

You could probably leave off the last if else statement and just use an else to show the error message, as it can be assumed that an error happened. Depends on how you want to handle the error from the image failing to load.

clhenrick (commented):
@mukhtyar now I'm getting an empty chart when selecting January where as before I think there was data displaying? Screenshot:

Screen Shot 2021-10-13 at 5 07 19 PM

Also the image is still not changng for me in static map. I tried in a different browser as well. Left some comments / suggestions.

src/components/tools/Location/StaticMap.svelte
mukhtyar: Thanks for updating this!

mukhtyar:
> 
> 
> @mukhtyar now I'm getting an empty chart when selecting January where as before I think there was data displaying? Screenshot:
> 
> Screen Shot 2021-10-13 at 5 07 19 PM
> 
> Also the image is still not changng for me in static map. I tried in a different browser as well. Left some comments / suggestions.

This should be now fixed by https://github.com/berkeley-gif/caladapt-website-2021/pull/84/commits/23a3c5fe825ae9e491a8722ce5b88e044d9b6b05

mukhtyar:
[Tools: CDD/HDD](https://trello.com/c/CAGsPasI/50-tools-cdd-hdd)

mukhtyar:
[Bug: CDD/HDD chart when freq=monthly & month=January](https://trello.com/c/e1qN3hKv/99-bug-cdd-hdd-chart-when-freqmonthly-monthjanuary)

PR #86, opened 2021-10-14
Add Embed component to handle iframing V2 tools

elehmer:
This adds a partial component Embed that creates iframe component to handle Cal-Adapt V2 tools. Adds one example for the streamflow tool.

### DIFF
--- a/src/partials/Embed.svelte
+++ b/src/partials/Embed.svelte
@@ -8,14 +8,12 @@
   let v2_frame;
 
   onMount(() => {
-    window.addEventListener(
-      "message",
-      (e) => {
-        // use message data that was passed from iframe page to set height
-        v2_frame.style.height = `${e.data.height}px`;
-      },
-      false
-    );
+    window.addEventListener("message", (e) => {
+      // use message data that was passed from iframe page to set height
+      if (typeof e.data.height !== "number")
+        return console.warn("postMessage from iframe not received!");
+      return (v2_frame.style.height = `${e.data.height}px`);
+    });
   });
 </script>
 
@@ -23,7 +21,7 @@
   iframe {
     width: 100vw;
     height: 100vw;
-    overflow: hidden;
+    overflow-y: hidden;
   }
 </style>
 
@@ -32,6 +30,4 @@
   title="{title}"
   frameborder="0"
   allow="fullscreen"
-  bind:this="{v2_frame}"
->
-</iframe>
+  bind:this="{v2_frame}"></iframe>
--- a/src/partials/Embed.svelte
+++ b/src/partials/Embed.svelte
@@ -1,0 +1,20 @@
+<script>
+  export let path;
+  export let title;
+
+  $: src = `https://v2.cal-adapt.org/${path}_iframe`;
+</script>
+
+<style>
+  .iframe {
+    width: 100vw;
+    height: 100vh;
+  }
+</style>
+
+<iframe
+  class="iframe"
+  src="{src}"
+  title="{title}"
+  frameborder="0"
+  allow="fullscreen"></iframe>
--- a/src/partials/Embed.svelte
+++ b/src/partials/Embed.svelte
@@ -10,9 +10,11 @@
   onMount(() => {
     window.addEventListener("message", (e) => {
       // use message data that was passed from iframe page to set height
-      if (typeof e.data.height !== "number")
-        return console.warn("postMessage from iframe not received!");
-      return (v2_frame.style.height = `${e.data.height}px`);
+      if (e.data.height && typeof e.data.height === "number") {
+        v2_frame.style.height = `${e.data.height}px`;
+      } else {
+        console.warn("postMessage from iframe not received!");
+      }
     });
   });
 </script>
--- a/src/partials/Embed.svelte
+++ b/src/partials/Embed.svelte
@@ -1,20 +1,37 @@
 <script>
+  import { onMount } from "svelte";
   export let path;
   export let title;
 
   $: src = `https://v2.cal-adapt.org/${path}_iframe`;
+
+  let v2_frame;
+
+  onMount(() => {
+    window.addEventListener(
+      "message",
+      (e) => {
+        // use message data that was passed from iframe page to set height
+        v2_frame.style.height = `${e.data.height}px`;
+      },
+      false
+    );
+  });
 </script>
 
 <style>
-  .iframe {
+  iframe {
     width: 100vw;
-    height: 100vh;
+    height: 100vw;
+    overflow: hidden;
   }
 </style>
 
 <iframe
-  class="iframe"
   src="{src}"
   title="{title}"
   frameborder="0"
-  allow="fullscreen"></iframe>
+  allow="fullscreen"
+  bind:this="{v2_frame}"
+>
+</iframe>
--- a/src/routes/tools/streamflow/index.svelte
+++ b/src/routes/tools/streamflow/index.svelte
@@ -22,17 +22,10 @@
   export let tool;
 </script>
 
-<style>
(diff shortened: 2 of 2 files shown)


src/routes/tools/streamflow/index.svelte
```diff
@@ -18,28 +18,22 @@
 
 <script>
   import { InlineNotification } from "carbon-components-svelte";
```
clhenrick: @elehmer you can remove this import as it's no longer needed

src/routes/tools/streamflow/index.svelte
```diff
@@ -18,28 +18,22 @@
...
-  import { Header } from "~/components/tools/Partials";
+  import Embed from "~/partials/Embed.svelte";
 
   export let tool;
 </script>
 
+<style>
+  .iframe-container {
```
clhenrick: @elehmer I think it would be better if we could set the height to the actual height of the content in the iframe so that it doesn't scroll vertically. [See this SO answer](https://stackoverflow.com/a/9163087) for one way of doing it. 

To get a reference to the `<iframe>` element you'll need to use Svelte's method of element binding. More [in the Svelte docs](https://svelte.dev/docs#bind_element).

clhenrick (changes requested):
@elehmer thanks for submitting this! Let me know if you can get the dynamic height solution working mentioned below. Happy to help / pair on it.

src/partials/Embed.svelte
```diff
@@ -0,0 +1,20 @@
+<script>
+  export let path;
+  export let title;
+
+  $: src = `https://v2.cal-adapt.org/${path}_iframe`;
+</script>
+
+<style>
+  .iframe {
```
clhenrick: p.s. @elehmer with Svelte you don't need to rely on classes when styling a component unless there are multiple instances of the same element. You can just do `iframe { ... }` and Svelte will scope the CSS to this component only.

clhenrick:
> @elehmer thanks for submitting this! Let me know if you can get the dynamic height solution working mentioned below. Happy to help / pair on it.

Although now I'm realizing that this will not work on cross-origin domains, so it might be hard to get working locally. You might need to write the code locally and then deploy to our server at a different subdomain like `test.cal-adapt.org` to test it.

elehmer:
I'm either running into cross-origin domain at the window level or it is reporting the wrong height for scrollHeight (probably because the iframe is loading async and is dynamically generated from code). I can't quite remember when cross-origin crops up. Won't we run into the cross-origin issue on the C-A server as well or will it be ok since the cal-adapt.org portion is the same?

I've tried using the onMount on both the Embed.svelte and index.svelte for the tool with the same results. I can set it to a fixed const and it works but if set too high the content is stretched in the iframe. Combining the fixed height with overflow hidden almost completely disables the scroll with just a slight scroll that isn't very noticeable.

Do we have a place to test the cross-origin issue and see if we can get the right value back?

Also, is there an event we can use that runs later then onMount that might return the right value?

mukhtyar:
@elehmer Let's add a script to the iframed versions of the tools that calculates the body height after rendering and use `postMessage` to communicate it to the iframe. Hopefully that should work with local testing. Otherwise we will have to set up another subdomain to test it.

clhenrick:
> @elehmer Let's add a script to the iframed versions of the tools that calculates the body height after rendering and use `postMessage` to communicate it to the iframe. Hopefully that should work with local testing. Otherwise we will have to set up another subdomain to test it.

@elehmer @mukhtyar maybe the easiest way to do it would be to add a resize event listener to the iframed page and have it call `postMessage` when it fires.

elehmer:
I have a working version now using the `postMessage` passing. Working out some annoying binding issues...

elehmer:
The iframe resizes for the streamflow tool using `postMessage`. Let me know if you see the same behavior I did. If so, I can modify the other v2 iframes to use the `postMessage`.

clhenrick:
thanks @elehmer I'll take another look

mukhtyar:
@elehmer it looks good to me. Tried it in a few browsers on a laptop and desktop.

src/partials/Embed.svelte
```diff
@@ -0,0 +1,37 @@
...
+  $: src = `https://v2.cal-adapt.org/${path}_iframe`;
+
+  let v2_frame;
+
+  onMount(() => {
+    window.addEventListener(
+      "message",
+      (e) => {
+        // use message data that was passed from iframe page to set height
+        v2_frame.style.height = `${e.data.height}px`;
```
clhenrick, line 14: it might be good to do a check in an `if` statement here to make sure `e.data.height` exists, and if not emit a `console.warn`

src/partials/Embed.svelte
```diff
@@ -0,0 +1,37 @@
...
+</style>
+
+<iframe
+  src="{src}"
+  title="{title}"
+  frameborder="0"
+  allow="fullscreen"
+  bind:this="{v2_frame}"
+>
+</iframe>
```
clhenrick: also @elehmer just FYI a minor style thing in Svelte: if an element has no children then you can self close it like:

```svelte
<iframe
  src="{src}"
  title="{title}"
  frameborder="0"
  allow="fullscreen"
  bind:this="{v2_frame}"
/>
```

src/routes/tools/streamflow/index.svelte
```diff
@@ -27,19 +26,6 @@
...
-        title="Under Construction"
-        subtitle=""
-      />
-    </p>
-  </div>
-</Header>
-
-<!-- placeholder div to add height -->
-<div id="explore"></div>
+<div>
```
clhenrick: @elehmer I don't think you need this `<div>` element here.

src/partials/Embed.svelte
clhenrick: For accessibility we should add a `title` attribute with a description of the embed. We could pass that in as a prop from the `tooname/index.svelte` parent.

clhenrick (commented):
@elehmer minor suggestions in addition to the solution we discovered of using `scrolling=no` on the `iframe` element.

src/partials/Embed.svelte
elehmer: Oh weird I had the iframe self closing but then fiddled with it in Sublime Text and it changed it lol. I'll fix that one.

clhenrick: No worries. There might be a Svelte plugin or something for Sublime Text. Also I realized you already had a title attribute, my bad.

elehmer:
Fixed the slight scrolling of iframe by adding `overflow-y: hidden` to the body of the v2 embedded html, so the code is over in that repo. Could also have been fixed by using  `scrolling=no` on iframe but that is deprecated so went with CSS option. Added error handling of `postMessage` data and cleaned up unneeded code. This should now be functional and can be applied to rest of the tools.

src/partials/Embed.svelte
```diff
@@ -0,0 +1,33 @@
...
+  export let title;
+
+  $: src = `https://v2.cal-adapt.org/${path}_iframe`;
+
+  let v2_frame;
+
+  onMount(() => {
+    window.addEventListener("message", (e) => {
+      // use message data that was passed from iframe page to set height
+      if (typeof e.data.height !== "number")
```
clhenrick: @elehmer I don't think there's a need for using `return` here. You can simplify this / make it a little more readable by making it an `if / else` statement. I might do the check like `if (e.data.height && typeof e.data.height === "number") { ... } else { ... }` in case `height` might not be there and lead with the truthy statement as it's generally easier to follow when something is truthy then fall back to falsey IMO.

clhenrick (approved):

PR #88, opened 2021-10-20
Streamflow QA / QC

elehmer:
This PR tries to fix the content running off the bottom of the iframe area. It adds 100 pixels to the `postMessage` returned height and sets the CSS of the iframe to display: block. The CSS margin: 0 auto was also added to the V2 streamflow iframe document to center it in the parent page. Seems to fix most issues but if the page is resized below the width of the iframe the header and footer of the parent page set to the page width not the iframe width so the iframe juts out to the right when you use the horizontal scrollbar.

### DIFF
--- a/src/partials/Embed.svelte
+++ b/src/partials/Embed.svelte
@@ -9,17 +9,21 @@
 
   onMount(() => {
     window.addEventListener("message", (e) => {
-      // use message data that was passed from iframe page to set height
-      if (e.data.height && typeof e.data.height === "number") {
-        v2_frame.style.height = `${e.data.height + 100}px`;
+      // use message data that was passed from iframe page to set height and width
+      if (e.data.height && e.data.width) {
+        if (
+          typeof e.data.height === "number" &&
+          typeof e.data.width === "number"
+        ) {
+          v2_frame.style.height = `${e.data.height + 100}px`;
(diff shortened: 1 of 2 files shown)


src/partials/Embed.svelte
```diff
@@ -23,7 +33,7 @@
   iframe {
     width: 100vw;
     height: 100vw;
-    overflow-y: hidden;
+    display: block;
```
clhenrick, line 30: also add `margin: 0 auto;`

clhenrick (commented):
@elehmer looks good, I think we actually should add a wrapper div element around the Embed instance and add `overflow-x: auto;` to it so that when the browser viewport width is less than the iframe width the content can at least scroll horizontally without expanding past the document body and making the footer & header look out of place.

src/partials/Embed.svelte
elehmer: Both added it latest commit with `margin: 0 auto` removed from body of iframe insert code.

clhenrick:
@elehmer I think the `width: 100vw` is creating unnecessary horizontal scrolling, at least when I'm on Firefox on MacOS. I made some temporary changes that fix the issue on my end, can you let me know what you see on your end? 

To clarify, I don't think we need the iframe's embedded content to be able to scroll horizontally since the tools on the v2 site don't do that anyway.

PR #89, opened 2021-10-22
Add wind extreme weather

mukhtyar:
Adds wind to the Extreme Weather tool and breaks up the _ExploreData component into smaller chunks.

### DIFF
--- a/src/components/tools/Charts/Observations/Circles.svelte
+++ b/src/components/tools/Charts/Observations/Circles.svelte
@@ -6,7 +6,6 @@
   const dispatch = createEventDispatcher();
 
   export let y1;
-  export let units;
 
   $: midHeight = $yScale.bandwidth() / 2;
 
@@ -49,7 +48,36 @@
 
 <g class="data">
   {#each $data as d}
-    {#if d.value}
+    {#if Array.isArray(d.value)}
+      {#each d.value as val, i}
+        <g>
+          <circle
+            cy="{$yGet(d) + midHeight}"
+            cx="{$xScale(val)}"
+            class="circle"
+            r="{5}"
+          >
+          </circle>
+          {#if i == 1}
+            <line
+              x1="{$xScale(val)}"
+              x2="{$xScale(val)}"
+              y1="{y1}"
+              y2="{$yGet(d) + midHeight}"
+              class="line"
+            >
+            </line>
+            <text
+              y="{$yGet(d) + midHeight + 4}"
+              x="{$xScale(i) + 10}"
+              class="label"
+            >
+              {d.valueLabel}
+            </text>
+          {/if}
+        </g>
+      {/each}
+    {:else if !isNaN(d.value)}
       <g>
         <line
           x1="{$xGet(d)}"
@@ -70,8 +98,7 @@
         >
         </circle>
         <text y="{$yGet(d) + midHeight + 4}" x="{$xGet(d) + 10}" class="label">
-          {d.value}
-          {units}
+          {d.valueLabel}
         </text>
       </g>
     {:else}
--- a/src/components/tools/Charts/Observations/Circles.svelte
+++ b/src/components/tools/Charts/Observations/Circles.svelte
@@ -2,7 +2,7 @@
   import { getContext, createEventDispatcher } from "svelte";
   import { raise } from "layercake";
 
-  const { data, xGet, yGet, xScale, yScale } = getContext("LayerCake");
+  const { data, xGet, yGet, yScale } = getContext("LayerCake");
   const dispatch = createEventDispatcher();
 
   export let y1;
@@ -48,36 +48,7 @@
 
 <g class="data">
   {#each $data as d}
-    {#if Array.isArray(d.value)}
-      {#each d.value as val, i}
-        <g>
-          <circle
-            cy="{$yGet(d) + midHeight}"
-            cx="{$xScale(val)}"
-            class="circle"
-            r="{5}"
-          >
-          </circle>
-          {#if i == 1}
-            <line
-              x1="{$xScale(val)}"
-              x2="{$xScale(val)}"
-              y1="{y1}"
-              y2="{$yGet(d) + midHeight}"
-              class="line"
-            >
-            </line>
-            <text
-              y="{$yGet(d) + midHeight + 4}"
-              x="{$xScale(i) + 10}"
-              class="label"
-            >
-              {d.valueLabel}
-            </text>
-          {/if}
-        </g>
-      {/each}
-    {:else if !isNaN(d.value)}
+    {#if !isNaN(d.value)}
       <g>
         <line
           x1="{$xGet(d)}"
@@ -103,7 +74,7 @@
       </g>
     {:else}
       <g>
-        <text y="{$yGet(d) + midHeight + 4}" x="{40}" class="label-na">
+        <text y="{$yGet(d) + midHeight - 5}" x="{45}" class="label-na">
           Data Not Available
         </text>
       </g>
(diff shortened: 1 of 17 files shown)


src/components/tools/Charts/Observations/Circles.svelte
```diff
@@ -49,7 +48,7 @@
 
 <g class="data">
   {#each $data as d}
-    {#if d.value}
+    {#if !isNaN(d.value)}
```
clhenrick, line 51: One thing I forgot about using `isNaN` for type checking is that some non-numeric data types will evaluate to `false` like `[]` or `{}` for example. But if you always expect `d.value` to either be a number or `null` or `undefined` then it's okay to use on its own. A more strict way (which I don't think needs to be used here) is to type check for a number first (`typeof d.value === "number"`) then use `!isNaN(d.value)`.

src/routes/tools/extreme-weather/_ExploreData.svelte
```diff
@@ -91,51 +50,29 @@
...
   } else {
-    isLoading = true;
+    histogramData = null;
   }
 
-  function updateDataExtent(arr) {
-    const dataExtent = extent(arr);
-    return [dataExtent[0] - 2, dataExtent[1] + 2];
-  }
+  afterUpdate(() => {
```
clhenrick: @mukhtyar curious if you found that you needed to use `afterUpdate` here to set the value for `chartTitle`? Or would `$:chartTitle = ` work? For wildfire I found I didn't need it, not sure what the issue is with snowpack where I did need it.

src/routes/tools/extreme-weather/_ExploreData.svelte
```diff
@@ -170,464 +109,76 @@
...
       ["station", $location.title],
       ["variable", $climvar.label],
       ["units", $climvar.units.imperial],
       ["record high", `{$baseline.high.value} on {$baseline.high.date}`],
       ["record low", `{$baseline.low.value} on {$baseline.low.date}`],
     ];
-    csvData = $baseline.values;
     printContainer = document.querySelector("#explore-data");
-    printSkipElements = ["settings", "chart-controls"];
+    printSkipElements = ["settings", "bx--tabs--container"];
```
clhenrick: I don't think the Extreme Weather uses tabs like Snowpack and Wildfire do?

src/routes/tools/extreme-weather/_ExtremeWeatherCharts.svelte
```diff
@@ -0,0 +1,254 @@
...
+      );
+    } finally {
+      isFetchingStore.set(false);
+    }
+  }
+</script>
+
+<div class="chart-controls">
+  {#if showRecentObs && !$isRecentObsLoaded}
+    <InlineLoading description="Fetching recent observations from NCEI..." />
```
clhenrick: IMO "loading" might be more appropriate for non-developers than "fetching", but not a big deal.

src/routes/tools/extreme-weather/_SettingsPanel.svelte
```diff
@@ -0,0 +1,144 @@
...
+
+  function showLearnMore({ slugs = [], content = "", header = "Glossary" }) {
+    dispatch("showLearnMore", { slugs, content, header });
+  }
+
+  function showChangeLocation() {
+    dispatch("showLoadLocation");
+  }
+
+  function changeClimvar(e) {
```
clhenrick: I would recommend breaking this out into two separate functions, `changeClimvar` and `extremesSideEffect` and calling them in another function `changeClimvarExtremes`.

src/routes/tools/extreme-weather/_constants.js
```diff
@@ -7,12 +7,36 @@ export const DEFAULT_STATION_LAYER = layers.find(
 );
 export const DEFAULT_STATION_ID = 11;
 
-export const CLIMATE_VARIABLES = ["tasmax", "tasmin"];
+export const CLIMATE_VARIABLES = ["tasmax", "tasmin", "wspeed"];
 export const DEFAULT_CLIMATE_VARIABLE = "tasmax";
+export const DEFAULT_SELECTED_EXTREME = "high";
+export const DEFAULT_SELECTED_DAY = new Date();
+export const DEFAULT_STATION = {
```
clhenrick, line 14: Should this data live somewhere in a shared constant? If not that's okay, I'm just thinking it could be helpful for future tools that might require a selected station.

clhenrick (approved):
Nice job refactoring @mukhtyar! I left a few small comments. I'm also noticing that Extreme Weather is missing the sections for Help and Resources, but guessing those will be updated eventually? 

## Here are a few visual things I spotted:

### Labels can crash with wind speed. 

One thing you might want to try for this is left and right aligning the text if they always display sequentially in the order p1, p10.

Screen Shot 2021-10-22 at 9 54 15 AM

### Weird formatting with values for confidence intervals in learn more for exceedance probability:

image

### Extreme event learn more text
The last line has an "is" which doesn't seem right:

image

src/routes/tools/extreme-weather/_constants.js
mukhtyar: The other 2 station datasets have different locations than these weather stations.

src/routes/tools/extreme-weather/_SettingsPanel.svelte
mukhtyar: @clhenrick Something like this?
```
  function changeClimvarExtremes({ detail }) {
    const { id } = detail;
    changeClimvar(id);
    extremesSideEffect(id)
  }

  function changeClimvar(id) {
    climvarStore.set(id);
    console.log("climvar change");
  }

  function extremesSideEffect(id) {
    if (id === "tasmax") {
      extremesStore.set("high");
    } else if (id === "tasmin") {
      extremesStore.set("low");
    } else {
      extremesStore.set("high");
    }
  }
```

src/routes/tools/extreme-weather/_ExploreData.svelte
mukhtyar: It works without `afterUpdate`

src/routes/tools/extreme-weather/_ExtremeWeatherCharts.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/89/commits/08e578e7d895f436b016bef365d98ae8be0c5ee4

src/routes/tools/extreme-weather/_ExploreData.svelte
mukhtyar: Right. Fixed it in https://github.com/berkeley-gif/caladapt-website-2021/pull/89/commits/92fa1994eaef5e3156173558ce870dbb5bb1e8ee

src/components/tools/Charts/Observations/Circles.svelte
mukhtyar: Good to know. In this case `d.value` will always be a `number` or `undefined` or `null`

src/routes/tools/extreme-weather/_SettingsPanel.svelte
mukhtyar: On second thoughts moved some of it into a constant:
```
  function changeClimvarExtremes({ detail }) {
    const { id } = detail;
    changeClimvar(id);
    extremesSideEffect(id)
  }

  function changeClimvar(id) {
    climvarStore.set(id);
  }

  function extremesSideEffect(id) {
    const { extremes } = DEFAULT_CLIMVAR_EXTREMES.find(({ climvar }) => id === climvar);
    extremesStore.set(extremes);
  }
```

clhenrick: @mukhtyar looks good to me.

PR #90, opened 2021-10-22
Add empty state for empty data array

mukhtyar:
Shows a blank LineArea chart if `data` array is empty

### DIFF
--- a/src/components/tools/Charts/LineArea/AxisY.svelte
+++ b/src/components/tools/Charts/LineArea/AxisY.svelte
@@ -18,6 +18,12 @@
   $: d = `M 0 0 L ${$width + $padding.left} 0`;
 </script>
 
+<style>
+  :global(.no-data .tick-label) {
+    display: none;
+  }
+</style>
+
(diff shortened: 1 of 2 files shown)


clhenrick:
@mukhtyar is there a way I can test this to see what it looks like?

mukhtyar:
image

clhenrick:
@mukhtyar I think it might help to have ticks for the y-axis as well if it's not too much effort. I guess the tricky part with that would be coming up with reasonable min and max values that would make sense for each dataset. If it's too complex to implement we could come back to it later.

mukhtyar:
@clhenrick yes coming up with reasonable min and max values for each dataset will be tricky. How about showing 5-10 y-axis ticks without any tick labels?

clhenrick:
@mukhtyar that sounds reasonable to me, maybe including the grid lines with the y-axis ticks as well? And perhaps the label for the y values?

mukhtyar:
@clhenrick how's this?  image

clhenrick:
@mukhtyar that looks good to me

PR #93, opened 2021-10-25
Update stats annual avg

mukhtyar:
Refactor dashboard stats panel `RangeAvg` component

### DIFF
--- a/src/components/tools/Stats/AvgRange.svelte
+++ b/src/components/tools/Stats/AvgRange.svelte
@@ -3,27 +3,33 @@
   import { extent, mean, merge } from "d3-array";
   import StatBlock from "./StatBlock.svelte";
 
-  // Array of data values grouped by date
-  // Each date has an array of 1 or more timeseries values
+  /** Array of data values grouped by date, e.g.
+  {
+    date:  Date Sat Dec 31 1949 16:00:00 GMT-0800 (Pacific Standard Time),
+    values: [
+      { id: "livneh", label: "Observed", value: 74.6},
+      { id: "CanESM2", label: "CanESM2 (Average)", value: 75.2},
+      { id: "CNRM-CM5", label: "CNRM-CM5 (Cool/Wet)", value: 73.3}
+    ]
+  }
+  */
   export let data;
 
-  // List of groups
-  // e.g. modeled historical, modeled projections, observed historical
+  /** List of 1 or more groups
+   * e.g. modeled historical, modeled projections, observed historical */
   export let groupList;
 
-  // List of periods
-  // e.g. baseline, mid-century, end-century
+  /** List of 1 or more periods
+  e.g. baseline, mid-century, end-century */
   export let periodList;
 
-  // Id of default group to initialize StatPanel
-  // Optional, defaults to first item in groupList
+  /** Id of default group to initialize StatPanel. This prop is optional. Defaults to first item in groupList */
   export let groupId;
 
-  // Id of default period to initialize StatPanel
-  // Optional, defaults to first item in periodList
+  /** Id of default period to initialize StatPanel. This prop is optional. Defaults to first item in periodList */
   export let periodId;
 
-  // List of selected models
+  /** List of selected models */
   export let models;
 
   export let format = (d) => d;
@@ -64,7 +70,7 @@
     // e.g. baseline, mid-century, end-century or a custom period
     const dataByPeriod = data.filter(subsetByYears(start, end));
     const values = merge(dataByPeriod.map(({ values }) => values));
-    // Filter data for selected period by selected group
+    // Filter values for selected group
     // e.g. modeled historical, modeled projections, observed historical
     let dataByGroup;
     // For modeled historical/projections, check if the value id is a model
--- a/src/components/tools/Stats/AvgRange.svelte
+++ b/src/components/tools/Stats/AvgRange.svelte
@@ -90,18 +90,20 @@
       ];
     } else {
       // For observed  historical, check if value id is not a model name
-      // and value id is not for the envelope. For now this value id = "livneh"
+      // and value id is not for the envelope. For now this would be
+      // the same as checking for id = "livneh"
       dataByGroup = values.filter(
         ({ id }) => !models.includes(id) && !id.includes("range")
       );
+      // Range is not displayed for observed data
+      return [
+        {
+          id: "average",
+          label: `${periodLength} YEAR AVG`,
+          value: calculateAverage(dataByGroup),
+        },
+      ];
     }
-    return [
-      {
-        id: "average",
-        label: `${periodLength} YEAR AVG`,
-        value: calculateAverage(dataByGroup),
-      },
-    ];
   }
 
   function update({ detail }) {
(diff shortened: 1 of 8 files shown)


src/components/tools/Stats/AvgRange.svelte
```diff
@@ -0,0 +1,134 @@
...
+
+  /** List of 1 or more groups
+   * e.g. modeled historical, modeled projections, observed historical */
+  export let groupList;
+
+  /** List of 1 or more periods
+  e.g. baseline, mid-century, end-century */
+  export let periodList;
+
+  /** Id of default group to initialize StatPanel. This prop is optional. Defaults to first item in groupList */
```
clhenrick, line 26: if these props should have default values then would it make sense to define their default values?

For example:

```js
export let groupId = "modeled-historical";
export let periodId = "baseline";
```

src/components/tools/Stats/AvgRange.svelte
```diff
@@ -0,0 +1,134 @@
...
+    return "-";
+  }
+
+  function calculateRange(values) {
+    if (Array.isArray(values) && values.length) {
+      const minmax = extent(values, (d) => d.value);
+      return `${format(minmax[0])}–${format(minmax[1])}`;
+    }
+    // Empty state
+    return "-";
```
clhenrick: Small typography suggestion, maybe use an en-dash (`–`) or em-dash (`—`) here instead of a hyphen.

src/components/tools/Stats/AvgRange.svelte
```diff
@@ -0,0 +1,134 @@
...
+
+  /** List of selected models */
+  export let models;
+
+  export let format = (d) => d;
+  export let units;
+
+  let metrics = [];
+
+  function subsetByYears(start, end) {
```
clhenrick, line 40: Nice use of functional programming!

src/components/tools/Stats/StatOptions.svelte
```diff
@@ -0,0 +1,125 @@
...
+
+  $: yearsList = range(dateRange[0], dateRange[1], 1);
+  $: items = yearsList.map((d) => ({ id: d, text: `${d}` }));
+  $: startYear_selectedIndex, updateLinkedList();
+
+  // If user confirms changes, dispatch change event with selected groupd and period
+  // If user has selected a custom period, create a new period object
+  function update() {
+    group = groupList.find(({ id }) => id === selectedGroup);
+    if (selectedPeriod === "custom") {
```
clhenrick, line 50: When I tried updating the years for Mid-Century to a custom range (2030-2060) I got an error in the console:

```
Uncaught TypeError: items[startYear_selectedIndex] is undefined
```

src/routes/tools/annual-averages/_ExploreData.svelte
```diff
@@ -188,33 +187,44 @@
...
+        <AvgRange
           units="{$climvar.units.imperial}"
-          data="{statsData}"
+          data="{dataByDate
+            ? dataByDate.filter((d) => d.date.getUTCFullYear() < 2006)
+            : null}"
           isHistorical="{true}"
-          series="{'historical'}"
-          period="{'baseline'}"
+          groupList="{DEFAULT_STAT_GROUPS.filter((d) => d.historical)}"
```
clhenrick, line 196: I'm not quite understanding why two historical groups are being passed in here while only one is displayed when the component renders. It looks like the `groupId` determines which is rendered, but then why pass two groups vs. one?

src/components/tools/Stats/StatOptions.svelte
```diff
@@ -0,0 +1,125 @@
...
+        titleText="End"
+        placeholder="Select end year"
+        items="{filteredItems}"
+      />
+    </div>
+  {/if}
+
+  <p style="margin-top:1.5rem;">
+    <strong
+      >Scientists recommend looking at a period of at least a few decades like a
```
clhenrick: We could say this as a little more succinctly, maybe "Climate scientists recommend a period length of at least three decades."  I would at least avoid the word "like" and use "such as" as "like" makes it sound less professional IMO.

src/components/tools/Stats/StatOptions.svelte
```diff
@@ -0,0 +1,125 @@
...
+      30 year period.</strong
+    >
+  </p>
+  <p>
+    Future climate projections express natural climate variability. If you
```
clhenrick: I think avoiding the word "you" here would be good to make it sound more professional. Perhaps: "Analyzing a only a few years of a future climate projection may result in misleading anomalies while analyzing a decades long time period will yield more probable future conditions."

src/routes/tools/annual-averages/_ExploreData.svelte
clhenrick: Okay I see now that in the options you can choose between historic observed and historic modeled.

src/components/tools/Stats/StatOptions.svelte
```diff
@@ -0,0 +1,125 @@
...
+<Modal
+  primaryButtonText="Confirm"
+  secondaryButtonText="Cancel"
+  on:click:button--secondary="{() => (open = false)}"
+  bind:open
+  modalHeading="Change Stats"
+  on:submit="{update}"
+  on:close="{() => dispatch('cancel')}"
+>
+  <h5>Select Group</h5>
```
clhenrick: For a11y purposes it would be good to make these an `h5` class instead of an actual `<h5>` as they are not living under an `<h4>`. Interestingly in Modals I've heard it argued that because they're a new context you can technically use a `<h1>` to start with if you need to use headings.

clhenrick (approved):
@mukhtyar on a cursory review it looks good to me. Nice way of breaking up the child components and having a higher order parent component that contains most of the business logic.

I did find one bug when trying to set a custom yearly range which I noted below.

src/components/tools/Stats/AvgRange.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/93/commits/61f0bec714e4bf6e28f57fe0d5d33ee1bc9de132

src/components/tools/Stats/AvgRange.svelte
mukhtyar: good idea. replaced with em-dashes in https://github.com/berkeley-gif/caladapt-website-2021/pull/93/commits/dd775543893f3ba3bad6bfc2f4c869b59a2d77aa

src/components/tools/Stats/StatOptions.svelte
mukhtyar: updated help text in https://github.com/berkeley-gif/caladapt-website-2021/pull/93/commits/4bd10e94c783aefd9e0b27978806c210039000a8

src/components/tools/Stats/StatOptions.svelte
mukhtyar: good to know! using `h5 class in https://github.com/berkeley-gif/caladapt-website-2021/pull/93/commits/18d9d1e254382ab835c3c57b0c47edd7d30dda33

src/components/tools/Stats/StatOptions.svelte
mukhtyar: @clhenrick not sure how to recreate this. The start and end years for the Mid-Century period is fixed. Were you trying to create a custom time period?

mukhtyar: @clhenrick never mind. I found the bug. I was getting the End year from the wrong array.

PR #98, opened 2021-10-28
Generalize V2 iframe Embed to Non-updated tools

elehmer:
Initial WIP branch. The slug for tools with underscore works well but there are issues with the vertical sizing of iframe on load. Some tools are fixed by resizing.

Content cut off on load but on resize is fixed. (Extended Drought, Extreme Prec,SLR Hourly)

Content cut off on load but doesn't get fixed on resize. (LCCS, SLR CalFlod 3d)

Maps of Project Change looks fine.

Extreme Prec also has very LARGE white area between main tool and About div.


clhenrick:
@elehmer try the following:

In the `Embed.svelte` modify the `if` statement so that it checks for `v2_frame` like:

```js
      if (
        v2_frame &&
        e.data &&
        typeof e.data.height === "number" &&
        !isNaN(e.data.height)
      ) {
        v2_frame.style.height = `${e.data.height}px`;
      } else {
        console.warn("Embed did not receive postMessage width value");
      }
```

For the embeds that aren't resizing correctly, I think there is some CSS causing elements to overflow outside of the document's body element. In your script try changing `document.body.scrollHeight` to `document.documentElement.scrollHeight ` and see if that helps. Otherwise we'll need to fix the CSS on those pages.

In Extreme Precipitation there's CSS making the container that houses the chart too tall, the class `.chart2` has `min-height: 60vh`. When I disable that rule in the regular tool at [https://cal-adapt.org/tools/extreme-precipitation/](https://cal-adapt.org/tools/extreme-precipitation/) nothing changes. Check with @mukhtyar to see why that rule is there and whether or not it can be removed.

clhenrick (commented):
See my comment above

clhenrick:
[Embed old tools in Iframes](https://trello.com/c/PL4v8cPW/100-embed-old-tools-in-iframes)

PR #99, opened 2021-10-29
Qa1 extreme weather wind addition

mukhtyar:


src/routes/tools/_common/helpers.js
```diff
@@ -149,3 +153,14 @@ export function formatDataForExport(_arr) {
...
+
+/**
+ * Converts degrees into the corresponding quadrant
+ * on a 16 point compass
+ * @param {number}
+ * @return {string}
+ */
+export function getCompassQuadrant(deg) {
+  const i = Math.round((deg % 360) / DEFAULT_COMPASS_QUADRANT_ANGLE);
+  return DEFAULT_COMPASS_QUADRANTS.get(i);
```
clhenrick: @mukhtyar if you're using the index to access the value then there's no reason to use a `Map`, you can just use an array.

clhenrick (approved):
LGTM. Only suggestion is that you can just use an array instead of a map for the wind speed directions since you're accessing them by their index anyway.

src/routes/tools/_common/helpers.js
mukhtyar: True! Fixed via https://github.com/berkeley-gif/caladapt-website-2021/pull/99/commits/3b40a43661a6814173413c914a36760ba9400742

PR #103, opened 2021-11-01
Line area chart align dates

mukhtyar:
The xaxis in the Line Area chart is a time scale. D3 adds xaxis ticks for the 1st of every year. The data in Annual Avg, Snowpack, Wildfire, & other tools that use the Line Area chart have data with dates at end of the year. I noticed that the values plotted in the chart & ticks don't align. This PR addresses this issue.

### DIFF
--- a/src/components/tools/Charts/LineArea/Area.svelte
+++ b/src/components/tools/Charts/LineArea/Area.svelte
@@ -5,6 +5,7 @@
   const { xScale, yScale } = getContext("LayerCake");
 
   export let series;
+  export let setDatetoYearStart = true;
 
   let show = true;
   const legendItems = getContext("Legend");
@@ -16,7 +17,13 @@
   });
 
   $: path = area()
-    .x((d) => $xScale(d.date))
+    .x((d) => {
+      if (setDatetoYearStart) {
+        return $xScale(new Date(Date.UTC(d.date.getUTCFullYear(), 0, 1)));
+      } else {
+        return $xScale(d.date);
+      }
+    })
     .y0((d) => $yScale(d.min))
     .y1((d) => $yScale(d.max));
 
--- a/src/components/tools/Charts/LineArea/Chart.svelte
+++ b/src/components/tools/Charts/LineArea/Chart.svelte
@@ -6,6 +6,7 @@
   import { min, max } from "d3-array";
   import { setContext } from "svelte";
   import { writable } from "svelte/store";
+  import { timeYear } from "d3-time";
 
   import Line from "./Line.svelte";
   import Area from "./Area.svelte";
@@ -34,6 +35,7 @@
     tickFormat: timeFormat("%Y"),
     units: "",
   };
+  export let setDatetoYearStart = true;
 
   let chartContainer;
   const legendItems = writable(null);
@@ -144,6 +146,7 @@
       xScale="{scaleTime()}"
       xDomain="{[xmin, xmax]}"
       yDomain="{[ymin, ymax]}"
+      xPadding="{[10, 10]}"
       data="{data}"
     >
       <Svg>
@@ -161,14 +164,14 @@
         <g class="area-group">
           {#if areaData}
             {#each areaData as area}
-              <Area series="{area}" />
+              <Area series="{area}" setDatetoYearStart="{setDatetoYearStart}" />
             {/each}
           {/if}
         </g>
         <g class="line-group">
           {#if lineData}
             {#each lineData as line}
-              <Line series="{line}" />
+              <Line series="{line}" setDatetoYearStart="{setDatetoYearStart}" />
             {/each}
           {/if}
         </g>
--- a/src/components/tools/Charts/LineArea/Chart.svelte
+++ b/src/components/tools/Charts/LineArea/Chart.svelte
@@ -65,8 +65,9 @@
     // Set X Domain
     xmin = min(data, (arr) => min(arr.values, (d) => d.date));
     xmax = max(data, (arr) => max(arr.values, (d) => d.date));
-    if (xAxis.baseValue === 0) {
-      xmin = xAxis.baseValue;
+    if (setDatetoYearStart) {
+      xmin = timeYear.floor(xmin);
+      xmax = timeYear.floor(xmax);
     }
 
     // Set Y Domain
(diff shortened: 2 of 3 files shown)


mukhtyar:
The 1980 HadGEM2-ES values is 74.3, which is plotted just beyond the tick mark for 1980
image

src/components/tools/Charts/LineArea/Area.svelte
```diff
@@ -16,7 +17,13 @@
   });
 
   $: path = area()
-    .x((d) => $xScale(d.date))
+    .x((d) => {
+      if (setDatetoYearStart) {
+        return $xScale(new Date(Date.UTC(d.date.getUTCFullYear(), 0, 1)));
```
clhenrick, line 22: @mukhtyar is there a reason why we should do this at the component level and not at the data parsing level immediately after fetching the data?

clhenrick (commented):
@mukhtyar just a question about where this date manipulation is happening.

src/components/tools/Charts/LineArea/Area.svelte
clhenrick: I'm trying to think if there's a situation where we wouldn't want the dates for each datum to be at the first of the year and would need them to be at the end of the year. I'd like to understand why the API is formatting dates this way in the first place.

mukhtyar: So far all our tools show data by year (whether is annual or monthly averages). So we can do this at the data parsing level also. I thought doing it at the component level gives flexibility for the future & we don't have to make any changes to individual tools. I think the API formats the date this way because that's the output from the pandas library when it aggregates data over time (e.g. annual average at end of calendar year versus annual average at end of water year should have different dates).

mukhtyar:
[Bug: Align date & ticks in Line Area chart](https://trello.com/c/99iUlzgA/113-bug-align-date-ticks-in-line-area-chart)

src/components/tools/Charts/LineArea/Area.svelte
clhenrick: I think what we want to avoid though is having duplicate code in all the charting components, so maybe having an abstraction of some kind that can be `import`ed into modules makes sense. But then we would still have to remember to do this for each chart component which seems like a potentially error prone approach.

I'm generally hesitant to make decisions based on hypothetical future scenarios (e.g. eventually having daily & hourly data) and wait to address those types of scenarios / problems until they're arriving. So in that sense I think the simplest way to solve it would be at the data parsing level. We can always modify that functionality later when it's necessary to do so.

mukhtyar: Sounds good. We can do it in the `_data.js` files in each of the tools.

clhenrick: Why not do it in `transformResponse` as to avoid having to duplicate it for each tool?

mukhtyar: I think that would work for all tools so far except the Extreme Heat tool. The Extreme Heat Tool uses the daily values to calculate different metrics by year - e.g. number of extreme heat days, heat waves, timing of extreme heat days, etc. Maybe we can add a flag in `transformResponse` which will reset the date by default but can be set to false for any tools that need the date info.

clhenrick: @mukhtyar given our discussion yesterday where we learned that we can modify the API param `freq=AS` to change the date to default to the first of January when aggregating by year, I feel like that would be the best approach. We will have to remember to do it in the API calls for the relevant tools, but it pushes the work to the server rather than the client which is a benefit.

mukhtyar: I don't think that approach will work for all tools. I'll create a card in trello or make a list here documenting how each tool is impacted.

mukhtyar: @clhenrick so with our current tools the only place using `freq=AS` would work to get data with dates in the right format is for Cooling Degree Days - Annual. We could add an extra `freq=AS` parameters for the annual endpoints to coerce the date into the correct format, but that won't work with any monthly endpoints.
<html><body>
<google-sheets-html-origin>

Tool | Example URL | Date Format from API
-- | -- | --
Snowpack | https://api.cal-adapt.org/api/series/swe_month_CNRM-CM5_rcp45/events/?imperial=true&g=Point(-119.9688%2038.9063)&months=4 | 2006-04-30T00:00:00Z
Annual Averages | https://api.cal-adapt.org/api/series/tasmax_year_models-min_historical/events/?imperial=true&g=Point(-121.4688%2038.5938) | 1950-12-31T00:00:00Z
Wildfire - Area Burned - Annual Simulation | https://api.cal-adapt.org/api/series/fire_year_CanESM2_rcp45_bau_mu/events/?imperial=false&g=Point(-119.1563%2037.2188) | 1953-12-31T00:00:00Z
Wildfire - Area Burned - Monthly Simulation | https://api.cal-adapt.org/api/series/fire_month_CanESM2_rcp45_bau_mu/events/?imperial=false&months=7&g=Point(-119.1563%2037.2188) | 1953-07-31T00:00:00Z
Cooling Degree Days/Heating Degree Days - Annual | https://api.cal-adapt.org/api/series/tasmax_day_CanESM2_rcp45/cdd/?imperial=true&g=Point(-121.4688%2038.5938)&freq=A&thresh=65 | 2006-12-31T00:00:00Z
Cooling Degree Days/Heating Degree Days - Monthly | https://api.cal-adapt.org/api/series/tasmax_day_CNRM-CM5_rcp45/cdd/?imperial=true&g=Point(-121.4688%2038.5938)&freq=M&thresh=65 | This gets monthly data which is filtered


</body>
</html>

clhenrick: Thanks very much for explaining this @mukhtyar it's super helpful. Let's chat more about it during our 1x1 today.

bkg: In case it's helpful, the `S` (start) modifier can be added to any frequency so `MS` should return a date like 5/1 instead of 5/31.

mukhtyar: @bkg thanks for the info. It was very helpful! We have decided to use `AS` or `MS` in any tools where it makes sense. In other tools we will update the dates in the `_data.js` files after fetching the data from the API and before adding it to the chart. So closing this PR without merging it in.

PR #105, opened 2021-11-03
Use new Banner partial in pages

clhenrick:
This PR attempts to standardize the usage of the Banner that appears at the top of most pages by adding a Banner partial component. 

Banner partial has the following features:
- [x] full bleed background image
- [x] full bleed background image for small viewports / mobile
- [x] background color
- [x] background gradient
- [x] overlay color & opacity for images
- [x] title text
- [x] subtitle text
- [x] slot for icons
- [x] slot for cta button
- [x] style adjustment for title text

Update the following pages to use the Banner partial:
- [x] Homepage
- [x] Data
- [x] Help
- [x] Events
- [x] About
- [x] Blog
- [x] Blog slug template
- [x] Event slug template


clhenrick:
[Top Level Layout](https://trello.com/c/DyweNO2K/54-top-level-layout)

src/partials/Banner.svelte
```diff
@@ -0,0 +1,100 @@
...
+
+  @media (max-width: 672px) {
+    .banner {
+      min-height: 14rem;
+      background-image: var(--banner-img-mobile, var(--banner-img), none);
+    }
+  }
+
+  .banner-content > * + * {
+    margin-top: var(--banner-child-v-spacing, 2.5rem);
```
clhenrick: This variable isn't set anywhere so it will always be `undefined` and default to 2.5rem.

src/partials/Banner.svelte
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  .banner > div {
```
clhenrick: Not sure why this rule is here?

clhenrick (commented):
It's amazing how much CSS this removes by centralizing it in a single component

src/routes/help/_categories.js
```diff
@@ -1,20 +1,20 @@
 const categories = [
   {
     slug: "get-started",
-    title: "Get Started",
-    text: "Learn how to get started with Cal-Adapt",
+    title: "Getting Started",
```
mukhtyar: Did we settle on Getting Started or Get Started? I don't remember. The phrase might be used in couple of different locations, like blog posts or events or Help sections.

src/partials/Banner.svelte
```diff
@@ -0,0 +1,93 @@
+<script>
```
mukhtyar, line 1: The banner partial looks great. Feels like it should cover all our pages  and useful for any future content. Would we use this partial in individual Tools?

src/partials/index.js
```diff
@@ -0,0 +1,13 @@
+export { default as BackToTop } from "./BackToTop.svelte";
```
mukhtyar, line 1: Moving all the partials to `index.js` is a good call!

mukhtyar (approved):
Looks good! Just left a few questions in the comments.

src/routes/help/_categories.js
clhenrick: Oh right, I think we ended up settling on "Get Started" though I feel like in certain places "Getting Started" seems to read a little better? I'll change it back to "Get Started"

src/partials/Banner.svelte
clhenrick: Ah right, I missed the usage of it in the Tools pages because they have their own Header partial. We could use it there as well.

clhenrick: @mukhtyar I spent a little bit of time fussing with swapping the Tool Header with the Banner. There are a few subtle differences that make it a bit tricky and require overrides but it could potentially work.

PR #106, opened 2021-11-03
Bug fix: degree days base temperature input

clhenrick:
Fixes the race condition that can occur when updating the base temperature setting using the keyboard.


clhenrick:
[CDD/HDD chart data gets out of sync](https://trello.com/c/4yRMkjKw/103-cdd-hdd-chart-data-gets-out-of-sync)

mukhtyar (approved):
Looks good to me.

PR #111, opened 2021-11-08
Align dates chart ticks

mukhtyar:
This aligns the x axis ticks in Line Area chart with the values for Annual Averages, Snowpack, Wildfire & Degree Days tools. It uses the `freq=AS` param in Annual Averages & Degree Days tools to get data from API with dates at start of year. For Snowpack (monthly data only) & Wildfire (annual & monthly) using the `freq` param does not work, since the data is monthly. For these the dates are reset in `fetchSeries` function.

### DIFF
--- a/src/components/tools/Charts/LineArea/Chart.svelte
+++ b/src/components/tools/Charts/LineArea/Chart.svelte
@@ -144,6 +144,7 @@
       xScale="{scaleTime()}"
       xDomain="{[xmin, xmax]}"
       yDomain="{[ymin, ymax]}"
+      xPadding="{[10, 10]}"
       data="{data}"
     >
       <Svg>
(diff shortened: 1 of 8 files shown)


clhenrick (commented):
@mukhtyar in your previous PR for this @bkg [mentioned](https://github.com/berkeley-gif/caladapt-website-2021/pull/103#discussion_r742246415):

> In case it's helpful, the S (start) modifier can be added to any frequency so MS should return a date like 5/1 instead of 5/31.

If this is the case could we use `freq=MS` when tools need to fetch monthly data?

mukhtyar:
@clhenrick We can use `freq=MS` to get data for start of month, e.g. get  `2006-04-01T00:00:00Z` instead of `2006-04-30T00:00:00Z`. However for the LineArea chart, since we are plotting any monthly data by year, we still need the date to be `2006-01-01T00:00:00Z` for ticks & marks to line up .

PR #114, opened 2021-11-15
Add check lineareachart single value

mukhtyar:
This PR adds a check to see if the difference in the y axis domain values for a LineArea chart are less than a tolerance value. This tolerance value is derived from the precision required for different datasets (default tolerance is 1.0). If all y axis values are very close together (i.e. `almostEqual`), the `ymax` value for the y axis domain will be reset to the tolerance value.


mukhtyar:
For the Decadal Wildfire Probability chart, this sets the y domain for all charts to [0, 1.0], even if values are not close together because the tolerance value defaults to 1.0
image
image

mukhtyar:
For Wildfire Area Burned & Snowpack charts, if all the data is 0 or close to 0, this avoid placing the 0 axis in the middle of the chart
image
image

src/components/tools/Charts/LineArea/Chart.svelte
```diff
@@ -79,6 +82,18 @@
       );
     }
 
+    // Derive a tolerance value using precision specified by user, defaults to 1
+    // Precision represents the number of digits past the decimal point
+    const tolerance = yAxis.precision
```
clhenrick: I'm not sure I understand this part; the value of `tolerance` will always be a string like "1.0" or "1.00" (when `precision` is 2). You're also passing this string to the `almostEqual` function when it looks like it expects a decimal value for the 3rd argument, not a string.

clhenrick (commented):
@mukhtyar nice solution! I think I understand the gist of it but am wondering if you've tried it using values other than `1` for `precision`.

src/components/tools/Charts/LineArea/Chart.svelte
mukhtyar: Right, it should be a float value and not a string value. But yes, the default float value will be `1.0`

mukhtyar:
@chris Good catch. Using a precision value other than 1 is not going to change the tolerance. It will still default to `1.0`. Instead we could pass in a `maxValue` in addition to `precision` & customize the top tick value
```
const tolerance = yAxis.precision ? +format(`.${yAxis.precision}f`)(yAxis.maxValue) : 1
```

src/components/tools/Charts/LineArea/Chart.svelte
clhenrick: So is the idea to change it to another float like 2.0, 1.5, etc?

clhenrick:
@mukhtyar maybe instead of `maxValue` it could be `maxDefault`, the former sort of implies that the y domain will never exceed that value which isn't the case. Also wondering if there's a situation where you might want to specify a `minDefault` value other than zero.

mukhtyar:
@clhenrick good idea. Instead of `baseValue` maybe we can have a `minDefault` and `maxDefault`. A chart showing minimum temperatures could go potentially below zero.

mukhtyar:
@clhenrick The role of `minDefault` & `maxDefault` as currently implemented is to adjust the y axis domain so we dont get charts with 1 or 2 ticks. 
- If `minDefault` prop is specified by user, the chart baseline will be `minDefault` value. This is used to set the baseline for some charts to be 0.
- If `maxDefault` prop is specified by user, it is used enforce a minimum number of ticks

I looked at the [Vega library](https://vega.github.io/vega/docs/scales/) for a different way to do this. Vega uses `domainMax`, `domainMin`, `zero` props to override the scale domain. To control axis ticks it uses `tickCount`, `tickMinStep` & other props to adjust number of ticks. However there is no easy way to tell D3 to produce a minimum number of ticks using `tickCount`. D3 treats `tickCount` as a suggestion. This is also an [issue in the Vega library](https://github.com/vega/vega/issues/1398).

I can think of 2 options:
- go with current implementation
- change implementation
  - add props `domainMin` & `domainMax`. If either of these props are present they will override the domain derived from data `[ymin, ymax] = [domainMin, domainMax]`
  - add `niceMax` prop, if this prop is present it will update domain max if `niceMax > ymax`

mukhtyar:
Reverting 2 commits dealing with placement of the y Axis label to include it in a separate PR.

src/components/tools/Charts/Scatter/Chart.svelte
```diff
@@ -65,6 +101,12 @@
     setContext("Legend", legendItems);
   }
 
+  function isEmptyData(_data) {
```
clhenrick: Pull out `isEmptyData` into a helper module and import it so that we're not rewriting it for each place it's being used.

src/components/tools/Charts/Scatter/Chart.svelte
```diff
@@ -41,19 +46,50 @@
   let ymin;
   let ymax;
 
+  let noData = false;
+
+  $: if (Array.isArray(data) && isEmptyData(data)) {
+    noData = true;
+    xmin = new Date(Date.UTC(1950, 0, 1));
+    xmax = new Date(Date.UTC(2099, 0, 1));
+    ymin = 10;
```
clhenrick: If we're passing in default values for the y-axis min and max, then it seems odd to hard code them to something else here when the data is empty.

src/components/tools/Charts/LineArea/Chart.svelte
```diff
@@ -60,27 +65,36 @@
...
     }
 
-    ymin = +yAxis.tickFormat(ymin);
-    ymax = +yAxis.tickFormat(ymax);
+    ymax = max(data, (arr) =>
+      max(arr.values, (d) => ("max" in d ? d.max : d.value))
+    );
+
+    // Check if ymin & ymax are almost equal
+    const tolerance =
```
clhenrick: I'm finding this logic to be a little confusing because of the semantics of what the `yAxis` properties `minDefault` and `maxDefault` values imply vs. how they're being used, and because `maxDefault` becomes the value of `tolerance` but only if it's defined. The `minDefault` is actually acting as an override for y-axis min value, not as a fallback which to me is what "default" implies, that if something doesn't exist or isn't overridden then use the default value. 

I think it might be easier to follow if 

1. The value for `tolerance` is a separate optional property in the `yAxis` prop.
2. The values of `minDefault` and `maxDefault` were used only when the `data` is empty 

It does seem correct that in the case of when the values of `ymin` and `ymax` are almost equal that the `tolerance` value could be used to override the value of `ymax` to avoid the "flat line" chart problem.

src/helpers/utilities.js
```diff
@@ -590,3 +590,13 @@ export const logSingleStore = (store, prev) => {
     }
   });
 };
+
+/**
+ * Check is two numbers are within the given tolerance of one another
+ * @param {a} number
+ * @param {b} number
+ * @param {tolerance} tolerance value
```
clhenrick: you could also add a returns type here to convey that this function returns a boolean value.

clhenrick (commented):
@mukhtyar left some thoughts, let me know if you'd like to discuss further.

src/components/tools/Charts/LineArea/Chart.svelte
mukhtyar: @clhenrick you are right that the semantics of the terms `minDefault` & `maxDefault` are confusing in this case. What about this idea instead:
 - add props `domainMin` & `domainMax`. If either of these props are present they will override the domain `ymin` & `ymax` derived from data
  - add `niceMax` prop. If this prop is present it will update domain ymax if `niceMax > ymax`

clhenrick: @mukhtyar good idea to look at how Vega handles this. I think using the `domainMin`, `domainMax`, and `niceMax` properties for the `yAxis` prop makes sense. How would `tolerance` be used here? Or would we abandon that approach of determining if the `ymin` and `ymax` are almost equal?

mukhtyar: @clhenrick yes we would get rid of the `tolerance` approach

clhenrick: @mukhtyar sounds good to me.

src/components/tools/Charts/LineArea/Chart.svelte
```diff
@@ -50,8 +55,8 @@
     noData = true;
     xmin = new Date(Date.UTC(1950, 0, 1));
     xmax = new Date(Date.UTC(2099, 0, 1));
-    ymin = 10;
-    ymax = 50;
+    ymin = 0;
```
mukhtyar: the values of `ymin` and ymax` dont' matter because we are not showing the yaxis labels. the `ymin` and `ymax` are used to show y axis lines in the empty chart

src/components/tools/Charts/LineArea/Chart.svelte
```diff
@@ -60,27 +65,52 @@
 
   $: if (Array.isArray(data) && !isEmptyData(data)) {
```
clhenrick: Just wondering if this should be an `else` after the first `if` or maybe a nested `if / else`:

```js
$: if (Array.isArray(data)) {
  if (isEmptyData(data) {
    // handle empty data
  } else {
    // set chart props
  }
}
```
But that might mess with the reactive statements, not sure. Alternatively the `isEmptyData` function could check to see if the `data` array is an array.

src/components/tools/Charts/Scatter/Chart.svelte
```diff
@@ -41,19 +46,66 @@
   let ymin;
   let ymax;
 
-  $: if (data) {
+  let noData = false;
+
+  $: if (Array.isArray(data) && isEmptyData(data)) {
+    noData = true;
+    xmin = new Date(Date.UTC(1950, 0, 1));
```
clhenrick: Wonder if we should have these as defaults somewhere, e.g.

```js
export const DEFAULT_X_MIN = new Date(Date.UTC(1950, 0, 1));
export const DEFAULT_X_MAX = new Date(Date.UTC(2099, 0, 1));
export const DEFAULT_Y_MIN = 0;
export const DEFAULT_Y_MAX = 10;
```
Or will they vary too much between charts?

clhenrick (approved):
I think this works nicely, I tried out the three tools you added the yAxis props too and the charts rendered as expected when the y domain's min and max were both at or close to zero.

src/components/tools/Charts/Scatter/Chart.svelte
mukhtyar: We can add them as constants. I don't think they will vary too much between charts. Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/114/commits/8df7666d3e0ba8ae33b9018c11b10f19b9d0382b

src/components/tools/Charts/LineArea/Chart.svelte
mukhtyar: Moved the check if `data` is array into the helper function in https://github.com/berkeley-gif/caladapt-website-2021/pull/114/commits/4de29c3ff081545cbd9336d1bc4e8f09de34b036

mukhtyar:
[LineArea Chart renders oddly when showing a single value](https://trello.com/c/op1EYQyk/118-linearea-chart-renders-oddly-when-showing-a-single-value)

mukhtyar:
[LineArea Chart renders outside of chart extent](https://trello.com/c/AYV2VmRR/120-linearea-chart-renders-outside-of-chart-extent)

PR #115, opened 2021-11-15
Add annotation component

mukhtyar:
sample annotation
image


src/components/tools/Charts/Shared/Annotation/Labels.svelte
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    return style;
+  };
+</script>
+
+<style>
+  .annotation-label {
+    position: absolute;
+    max-width: var(--max-width, 200px);
```
clhenrick, line 62: where are the CSS variables being set? I don't see how they're meant to be overridden.

src/components/tools/Charts/LineArea/Chart.svelte
```diff
@@ -34,6 +36,8 @@
     tickFormat: timeFormat("%Y"),
     units: "",
   };
+  //export let annotations;
```
clhenrick: Did you mean to make `annotations` an exportable property? Otherwise this would mean every LineArea chart would have sample annotations applied, correct?

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
```diff
@@ -0,0 +1,162 @@
...
+
+  $: setStyle = (styles) => {
+    return Object.entries(styles)
+      .map(([key, value]) => `--${key}:${value}`)
+      .join(";");
+  };
+</script>
+
+<style>
+  .annotation-connector path {
```
clhenrick, line 166: same with these CSS variables, how can they be overridden? I assume there should be a way to pass down properties from the Chart component that sets them?

clhenrick (commented):
@mukhtyar I think the way you currently have it written each LineArea chart would get the default / dummy annotations as they're not override-able via a prop on the LineArea/Chart component.

I see CSS custom properties but no way to set them. If you want to use CSS custom properties in a Svelte component you need to set them via the component's props ([read more here](https://svelte.dev/docs#style_props)).

One other thing worth mentioning is if you're borrowing someone else's code it's a good practice to credit them in the code comments somewhere. It can also be helpful when referring back to this code in the future to know where the original code came from.

src/components/tools/Charts/LineArea/Chart.svelte
mukhtyar: `annotations` would be an exportable property. The `getSampleAnnotations` is just for testing.

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
```diff
@@ -0,0 +1,311 @@
...
+        {
+          type: "threshold",
+          orientation: "v",
+          source: {
+            anchor: "left-middle",
+            dx: 0,
+            dy: 10,
+          },
+          target: {},
+          style: { "stroke-dasharray": "6,4" },
```
mukhtyar, line 360: @clhenrick here's an example of how you'd pass in a custom style. this will override the default `stroke-dasharray` style of `0`

src/components/tools/Charts/Shared/Annotation/Labels.svelte
mukhtyar: @clhenrick I'll add in the reference to the Layercake code I'm using. The default props can be overriden by passing in a `styles` object with the css props like `style: { "stroke-dasharray": "6,4", "fill":"red" }`. Just realized this works with the `Connector` component, need to add it to the `Labels` component

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
clhenrick: That example is not setting a CSS custom property though. You would need to do something like:

```
style: {
  "--my-custom-prop": "some-value"
}
```

Is that how they are intended to be set?

clhenrick: Nevermind, I'm seeing that the `$: setStyle` reactive function expression is prefixing the property names with "--". I think it could just be a regular function, not sure it needs to be reactive?

src/components/tools/Charts/LineArea/Chart.svelte
clhenrick: Thanks for explaining.

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
mukhtyar: The child components of the chart run twice on page load, I can't figure out why. The first time they run the `height` and `width` of the chart are not the same as the final values, so the labels & connectors get put in the wrong position when using `data` to position the elements. Making the functions reactive recalculated the positions with correct `height` & `width` values

clhenrick: Got it, I didn’t even know you could make a function a reactive variable/statement in svelte.

src/components/tools/Charts/Shared/Annotation/Annotations.svelte
```diff
@@ -0,0 +1,26 @@
+<script>
+  // Annotation component based on annotated column example from LayerCake
+  // https://layercake.graphics/example/Column
+  import { getContext, afterUpdate, onMount } from "svelte";
```
clhenrick: a couple of these imports aren't used so you could remove them

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
```diff
@@ -0,0 +1,229 @@
...
+    : [];
+
+  // Creates a svg path for each arrow connector
+  $: setArrowPath = (anno, i, connector) => {
+    if (!Array.isArray(labelEls) || !labelEls.length) return;
+
+    // Get corresponding label div & bounding box
+    const label = labelEls[i];
+
+    const { type, source, target } = connector;
```
clhenrick: would `connector` ever not exist? If so it might be good to have a default value for it in the function params or here via `const { type, source, target } = connector || {}`

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
```diff
@@ -0,0 +1,229 @@
...
+
+    // If no type or source is specified do not create connector
+    if (!type || !source || !target) {
+      console.warn(
+        "cannot create arrow connector: missing required type or source or target prop"
+      );
+      return;
+    }
+
+    let sourceCoords;
```
clhenrick: You could pull out the logic used to define `sourceCoords` and `targetCoords` in separate functions and pass in `source` and `target` respectively.

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
```diff
@@ -0,0 +1,229 @@
...
+        dy: targetDy,
+      });
+    } else {
+      console.warn("cannot create connector: missing data prop for target");
+      return;
+    }
+
+    // Default to clockwise direction for swoop
+    const clockwise =
+      connector.clockwise === "undefined" ? true : connector.clockwise;
```
clhenrick: would it make more sense to just assume that `connector.clockwise` is a boolean here? In the annotations data is it actually getting set to a string "undefined" or should this be `typeof connector.clockwise === "undefined"`?

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
```diff
@@ -0,0 +1,229 @@
...
+
+    // If no source is specified do not create connector
+    if (!source) {
+      console.warn(
+        "cannot create threshold connector: missing required source prop"
+      );
+      return;
+    }
+
+    let sourceCoords;
```
clhenrick: Ditto here, this could be pulled out into a separate function. Sometimes I think it makes things a bit easier to follow in a long code block, e.g. `const sourceCoords = getSourceCoords(source)`;

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
```diff
@@ -0,0 +1,229 @@
...
+    } else {
+      console.warn(
+        "cannot create threshold connector: missing data prop for source"
+      );
+      return;
+    }
+
+    // Default to vertical orientation for threshold
+    const orientation =
+      typeof connector.orientation === "undefined"
```
clhenrick, line 142: This makes me think you meant  to use `typeof` above in a similar code block.

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
```diff
@@ -0,0 +1,229 @@
...
+    markerWidth="17"
+    markerHeight="17"
+    orient="auto"
+  >
+    <path d="M-6,-6 L 0,0 L -6,6"></path>
+  </marker>
+</defs>
+
+<g bind:this="{container}">
+  {#if Array.isArray(labelEls) && labelEls.length}
```
clhenrick, line 191: I wish svelte would let you create more than one component in a single file, if so this wouldn't have to be so nested. Seems like the only alternative is to use functions that return template literals and render them using `@html`.

src/components/tools/Charts/Shared/Annotation/Labels.svelte
```diff
@@ -0,0 +1,83 @@
...
+  export let annotations;
+  export let xGet;
+  export let yGet;
+
+  const vals = ["top", "right", "bottom", "left"];
+
+  $: labels = annotations.map((d) => d.label);
+
+  $: setPositionAndStyle = (label) => {
+    const { position, data, style, dx, dy } = label;
```
clhenrick: If there's any chance that `label` will be `undefined` it would be good to use a default param value.

src/components/tools/Charts/Shared/Annotation/Labels.svelte
```diff
@@ -0,0 +1,83 @@
...
+      // values are in pixel values or percentages
+      vals.forEach((val) => {
+        if (position[val]) {
+          inlineStyle += `${val}: ${position[val]};`;
+        }
+      });
+    } else if (data) {
+      // data object is a row from the data displayed in chart
+      // use the accessor functions to derive position
+      let xPos = xGet(data) ? xGet(data) : 0;
```
clhenrick, line 30: if `xGet()` and `yGet()` are returning a number or `undefined`/`null` than you could simplify this to `xGet(data) || 0`.

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
```diff
@@ -0,0 +1,458 @@
...
+/* --------------------------------------------
+ * parseCssValue
+ *
+ * Parse various inputs and return then as a number
+ * Can be a number, which will return the input value
+ * A percentage, which will take the percent of the appropriate dimentions
+ * A pixel value, which will parse as a number
+ *
+ */
+export function parseCssValue(d, i, width, height) {
```
clhenrick: Sort of nit picky but I feel like `i` isn't a good param name here, seems like it should be a boolean maybe called `orientVertical` that when true uses the value of `height` and when false uses `width`. The params `d` and `i` made me think this was being passed as an accessor function somewhere.

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
```diff
@@ -0,0 +1,458 @@
...
+/* --------------------------------------------
+ * getPositionFromAnchor
+ *
+ * Use label position & anchor to calculate
+ * source coordinates
+ *
+ */
+export function getPositionFromAnchor({ anchor, label, dx = 0, dy = 0 }) {
+  const elPos = getElPosition(label);
+  const [x, y] = anchor.split("-").map((q, j) => {
```
clhenrick: Also a little nit picky but using `q` and `j` here when usually `d` and `i` are used feels odd. If this was nested in another loop that already used `d` and `i` it would make more sense.

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
```diff
@@ -0,0 +1,458 @@
...
+    const h = hypotenuse(data[1][0] - data[0][0], data[1][1] - data[0][1]);
+
+    // get the distance at which chord of height h subtends {angle} radians
+    const d = h / (2 * Math.tan(angle / 2));
+
+    // get the radius {r} of the circumscribed circle
+    const r = hypotenuse(d, h / 2);
+
+    /*
+    SECOND, compose the corresponding SVG arc.
```
clhenrick: Not sure this comment is necessary, it's huge and we can always look up the syntax of the SVG path's `d` attribute.

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
```diff
@@ -0,0 +1,458 @@
...
+  };
+
+  render.y = function renderY(_) {
+    if (!arguments.length) return yValue;
+    yValue = _;
+    return render;
+  };
+
+  return render;
+}
```
clhenrick, line 275: More a comment and not suggesting you do this, but this is sort of the older style of writing modular JS from [Bostock's reusable chart pattern circa 2012](https://bost.ocks.org/mike/chart/) and JS has come a long way since. If it were me, I would write this as an ES6 class (maybe called `Arrow`) and each of these arrows would `extend` that class adding custom functionality. It just seems like a ripe opportunity for using object oriented programming.

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
```diff
@@ -0,0 +1,458 @@
...
+ * @typedef {Object} Target
+ * @property {Object} [data] – A row from data used in chart, used by layercake accessor functions to get x & y value
+ * @property {String} [x] - x position as number, pixel value or percentage
+ * @property {String} [y] - y position as number, pixel value or percentage
+ */
+
+/**
+ * The Data object, used by the Label & Connector objects
+ * @typedef {Object} Data
+ * @property {Date} [data.date = 0] - example for LineArea chart
```
clhenrick: would `data.date` actually be `0` here or a date object?

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
```diff
@@ -0,0 +1,458 @@
...
+
+/**
+ * The Position object, used by the Label object for absolute positioning
+ * @typedef {Object} Position
+ * @property {String} [position.top] - top edge as number, pixel value or percentage
+ * @property {String} [position.bottom] - bottom edge as number, pixel value or percentage
+ * @property {String} [position.left] - left edge as number, pixel value or percentage
+ * @property {String} [position.right] - right edge as number, pixel value or percentage
+ */
+export function getSampleAnnotations() {
```
clhenrick, line 344: maybe add a line break here so that it's clear that the type definition for Position isn't describing the function below it.

clhenrick (approved):
@mukhtyar nice work, this is a complex problem to solve and this seems like a solid approach to solving it. Really wish we had Story Book set up to showcase it. 

Left some minor comments and suggestions on the code, feel free to take them or leave them. 

For the demo I see that there are halos around the labels to help the text stand out which is great. I'm wondering if we might want a halo around the arrows as well. I didn't see the smaller swoopy arrow at first as it blends in with the other lines a little bit. One way of solving this would be to draw two arrows on top of each other and to give the bottom one a slightly wider stroke with a white stroke color. This could be an enhancement later though, don't think we need to add it now.

image

src/components/tools/Charts/Shared/Annotation/Labels.svelte
clhenrick: @mukhtyar forgot to mention that the font-size of 1rem feels a little too large when compared to the rest of the chart's labels; bumping it down to 0.875rem or 0.75rem might help it not clash so much with the axis tick labels for example.

src/components/tools/Charts/Shared/Annotation/Annotations.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/115/commits/641a71fdb94b28519a74dda97d05633c1c9df4fc

src/components/tools/Charts/Shared/Annotation/Connectors.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/115/commits/94e802dc819cba391425df0f4f0f6dea2fc97af1

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/115/commits/c966befb3c3363d537835c28d41aaf55f3462b44

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/115/commits/dab51a557f1598c7ef9b235e4e294954aeaa486b

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/115/commits/4c9240f1d60907c9986f65169782700f7c2651f1

src/components/tools/Charts/Shared/Annotation/annotationUtils.js
mukhtyar: fixed in  https://github.com/berkeley-gif/caladapt-website-2021/pull/115/commits/b672be7e3e2802f20f7d7ef81998018b313a090a

PR #126, opened 2021-11-24
Bug fix: help/glossary internal links

clhenrick:
- [x] add page anchor ids to all glossary terms
- [x] use page anchor links in terms when linking to other terms
- [x] adjust scroll position so that the viewport is at the correct height after clicking on an internal link to another term


mukhtyar (approved):
There is 1 additional place in the content which links to the climate normal entry in `content\faqs\what-baseline-time-period-should-i-use-to-compare-future-climate-projections.md`

clhenrick:
Thanks @mukhtyar good catch

clhenrick:
[Help: Glossary](https://trello.com/c/6wQS7SDZ/18-help-glossary)

PR #131, opened 2021-11-30
Extreme weather remove lowextremes wind

mukhtyar:


src/routes/tools/extreme-weather/_SettingsPanel.svelte
```diff
@@ -32,6 +32,9 @@
   const dispatch = createEventDispatcher();
   const { location } = locationStore;
 
+  let extremesList;
```
mukhtyar: set default value

src/routes/tools/extreme-weather/_constants.js
```diff
@@ -107,12 +107,13 @@ export const DOY_DESCRIPTION = `<p>This can be any day of the year you wish to s
 
 export const EXTREMES_DESCRIPTION = `<h5>High Extremes</h5>
```
clhenrick: @mukhtyar just remember that it's not good for a11y to use heading level elements to size text. Since it's about the same size as the regular `<p>` you could probably just do `<p><strong>High Extremes</strong></p>` for each of these.

clhenrick (approved):
@mukhtyar nice solution. 

Not related to this PR but I think currently the extreme selection title in the Settings Panel reads a little awkwardly. Could we maybe label it as "Select extreme type"?

src/routes/tools/extreme-weather/_constants.js
mukhtyar: Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/131/commits/b3890f612caa393a20aa0de797609c6311a64703

PR #133, opened 2021-12-01
Bug fixes: Change Location widget location & boundary selection 

clhenrick:
Addresses a couple issues with the Set Location widget: 

- [x] when a user changes the boundary type to place, instead of using the current location's center to select the nearest place geometry, it defaults to Acalanes Ridge, CA.
- [x] when clicking on the map to select a boundary type, the boundary that ends up being selected may not be the one the user clicked on.
- [x] potential network request failures when attempting to change the current location
- [x] missing title property for locagrid locations when changing boundary type from non-loca to loca


clhenrick:
Noting that this also may be an API issue, it looks like when an API call is made for /place/ and the distance_to param is used, results are being returned in alphabetical order, not in the order of the distance to the Point geometry in the distance_to param. 

Here's an example API call:
https://api.cal-adapt.org/api/place/?srs=4326&precision=4&distance_to=POINT(-120.7813%2038.7188)

clhenrick:
[Bug: When changing location boundary type the current map center coords are not used](https://trello.com/c/QFkjsZOr/110-bug-when-changing-location-boundary-type-the-current-map-center-coords-are-not-used)

src/components/tools/Partials/ChangeLocation.svelte
```diff
@@ -70,7 +70,7 @@
 
   async function updateBoundary(e) {
     if (!e.detail) return;
-    const { lng, lat } = currentLoc.center;
+    const [lng, lat] = currentLoc.center || [];
```
clhenrick: Maybe instead of `[]` this could be the default location center coords.

mukhtyar: @clhenrick  Yes that's correct. `currentLoc.center` is an array and not an object. Thats one bug. Another issue is whether `distance_to` param is supported for boundaries. See comment below.

mukhtyar: @bkg   the `distance_to` param is supported for station datasets, is it also supported for boundaries?  

Station example - https://api.cal-adapt.org/api/hadisdstations/?distance_to=POINT(-124+40)
this works, a `distance` property is added to `properties` for each feature

Boundary example -https://api.cal-adapt.org/api/place/?srs=4326&precision=4&distance_to=POINT(-120.7813%2038.7188)
 this does not add a `distance` property to each feature

bkg: I followed up on Trello, but currently no, `distance_to` is only available for HadISD stations. I suppose it's only a problem for any sparse boundary layers like places since a simple intersect will return a result for all the others.

src/components/tools/Partials/ChangeLocation.svelte
```diff
@@ -35,9 +35,15 @@
   let showSuggestions = false;
 
   async function mapClick(e) {
-    const addresses = await reverseGeocode(`${e.detail[0]}, ${e.detail[1]}`);
-    const feature = addresses.features[0];
-    currentLoc = await getFeature(feature, currentBoundary.id);
+    let newLocation;
```
clhenrick: @mukhtyar from what I can tell the call to `reverseGeocode` here isn't needed, since ultimately a lng, lat coordinate is being passed to `getBoundaryPolygon` (in `/helpers/geocode/` to find the appropriate geometry for the selected boundary type. Let me know if there's a case I'm missing.

clhenrick:
[Bug: Change Location problem with watershed boundary selection in Emigrant Wilderness](https://trello.com/c/wFECYMq4/122-bug-change-location-problem-with-watershed-boundary-selection-in-emigrant-wilderness)

src/components/tools/Partials/ChangeLocation.svelte
clhenrick: Edit: the one place where this changes things is that when a user selects a "place" boundary type when they click outside of a place polygon nothing will happen, which seems fine IMO.

mukhtyar: The place name returned by `reverseGeocode`  is used as the default location title if the user is using LOCA boundary. See functions `formatFeature` and `getTitle`. If user has selected a boundary other than LOCA, the location title is derived from the boundary polygon properties returned by API. But the `reverseGeocode` lookup could be moved so it only happens for LOCA boundaries.

clhenrick: I'm thinking that perhaps getting a place name for the title should be separate from selecting the grid cell. I'm noticing an odd behavior where sometimes clicking on a grid cell doesn't actually select it, but will select a neighboring grid cell. I believe this is because for certain cases the lat lon gets reverse geocoded to a neighboring place and thus neighboring grid cell when the grid cell being clicked on doesn't have a place associated with it in the Mapbox geocoding API.

Perhaps it makes sense to pass the coordinates from the map click directly to `getFeature` and then do the reverse geocoding once the user has confirmed their selection?

mukhtyar: That makes sense to me.

mukhtyar (approved):

clhenrick:
Note: waiting on an API update from @bkg to be completed before continuing with the bug fix for places boundary type.

PR #134, opened 2021-12-02
Bugfix update stats data change

mukhtyar:


mukhtyar:
[Stats Panel should re-render when data changes without data being set to null](https://trello.com/c/7uQzStT3/121-stats-panel-should-re-render-when-data-changes-without-data-being-set-to-null)

PR #145, opened 2021-12-03
Feature: add exception tracking via GA

clhenrick:
Use Google Analytics to log errors in places where things may fail, such as in async operations like network requests and tool UI updates. This will only be enabled for production deploys so that we aren't logging events during development or on the dev server URL.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -24,4 +24,7 @@ module.exports = {
   settings: {
     "svelte3/ignore-styles": () => true,
   },
+  globals: {
+    ga: "readonly",
+  },
 };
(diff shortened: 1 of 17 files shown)


mukhtyar (approved):

clhenrick:
[Log errors using GA event capturing](https://trello.com/c/0awtd4YV/134-log-errors-using-ga-event-capturing)

clhenrick:
@mukhtyar FYI I was able to verify that this is working using by:

1. creating a production deploy build via `npm run deploy-prod -- --transfer=false`
2. running the build locally via `npx serve __sapper__/export`
3. Opening the Chrome browser and using the [Google Tag Assistant](https://tagassistant.google.com/) to make sure an exception was being logged by throwing in error in the source code.

Similarly, I was able to verify that exceptions **are not** being logged when running `npm run dev` and `npm run deploy-dev`.

PR #146, opened 2021-12-03
return existing threshold list instead of undefined

mukhtyar:


clhenrick (approved):
LGTM

mukhtyar:
[Bug: Navigating away and back to Extreme Heat tool throw error](https://trello.com/c/ZWPfftYJ/132-bug-navigating-away-and-back-to-extreme-heat-tool-throw-error)

PR #147, opened 2021-12-04
Use single component for change location / station

clhenrick:
Merges the separate components from the Change Location and Change Station widgets into a single component so that bug fixes from #133 can be applied to both. This should be merged prior to #133


mukhtyar (approved):
Looks good!

src/components/tools/Partials/ChangeLocationStation.svelte
```diff
@@ -113,16 +134,17 @@
...
     dispatch("change", {
-      boundaryId: currentBoundary.id,
+      ...(currentBoundary && { boundaryId: currentBoundary.id }),
       location: currentLoc,
     });
   }
 
   function cancel() {
     currentLoc = location;
-    currentBoundary = boundary;
```
clhenrick, line 123: This should not be removed

PR #148, opened 2021-12-07
Bugfix fix Card Container alignment on Get Started page

mukhtyar:


src/components/cards/CardsContainer.svelte
```diff
@@ -23,6 +23,9 @@
   }
 </style>
 
-<ul style="--gap:{gridGap}rem; --card-width:{cardWidth}rem">
+<ul
+  class="card--list--container"
```
mukhtyar: added a class here to target the `ul` element containg cards in `_carbon-overrides.scss`

src/scss/site/_carbon-overrides.scss
```diff
@@ -46,7 +46,7 @@
 }
 
 .content {
-  ul:not(.bx--tabs__nav),
+  ul:not(.bx--tabs__nav):not(.card--list--container),
```
mukhtyar: This removes `padding-left` for the `ul` element containing the cards. Removing  `padding-left` makes the cards align with the header on all viewport sizes

src/components/cards/CardsContainer.svelte
clhenrick: @mukhtyar could we make the class name more similar to the Component name? Maybe `cards-container-list`? That way it's a little easier to find what it relates to in the Carbon style overrides `.scss` partial.

clhenrick (commented):
@mukhtyar thanks for debugging this. I just added a suggestion to change the class name you added to the `CardsContainer`'s `<ul>`

src/components/cards/CardsContainer.svelte
mukhtyar: updated class name in https://github.com/berkeley-gif/caladapt-website-2021/pull/148/commits/a9a6094666d89e45b8e526a6843f71572bb55959

PR #160, opened 2021-12-10
Break up Extreme Heat ExploreData into smaller components used in other tools

mukhtyar:


src/components/tools/Stats/AvgRange.svelte
```diff
@@ -135,7 +135,6 @@
   <StatBlock
     on:update="{update}"
     units="{units}"
-    data="{data}"
     groupList="{groupList}"
```
mukhtyar, line 138: Removed unused `data` prop

src/routes/tools/extreme-heat/_SettingsPanel.svelte
```diff
@@ -0,0 +1,174 @@
...
+    thresholdListStore.add(e.detail);
+  }
+</script>
+
+<!-- Chart Settings -->
+<div class="block">
+  <span class="bx--label">Select Location</span>
+  <StaticMap
+    location="{$location}"
+    width="{350}"
```
mukhtyar: remove `width`. set `height` to 250

src/routes/tools/extreme-heat/_ChartTitle.svelte
```diff
@@ -0,0 +1,45 @@
...
+  icon="{Location16}"
+  kind="ghost"
+  on:click="{loadLocation}"
+>
+  Change Location
+</Button>
+<h2 class="h3">
+  {title}
+</h2>
+<div class="h4">
```
clhenrick: I think using a `<p>` element for the `<div class="h4">` would be more semantic. Generally it's good to avoid having text that's only surrounded by a `<div>` element.

clhenrick (approved):
@mukhtyar nice work on splitting this up into separate components. 

I noticed a chart issue that's probably unrelated to this PR when I select Extreme Heat, timing indicator, for RCP8.5, and threshold of 65˚F, for a census track in Modesto:

Screen Shot 2021-12-10 at 3 40 37 PM

src/routes/tools/extreme-heat/_ChartTitle.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/160/commits/5c4cc241b846de4263d9a7eef65b5d15e0ef3293

PR #163, opened 2021-12-20
Add tool extreme precipitation

mukhtyar:


clhenrick:

src/components/tools/Charts/DotAndWhisker/DotAndWhisker.svelte
```diff
@@ -31,16 +31,16 @@
     <g class="{group}" transform="translate({$rScale(group)},0)">
       {#each values as d}
         <g
-          class="{d.key}"
+          class="{d.id}"
```
mukhtyar, line 34: Need to show/hide series when legend is clicked.

src/routes/tools/extreme-precipitation/_ChartTitle.svelte
```diff
@@ -0,0 +1,51 @@
...
+  on:click="{loadLocation}"
+>
+  Change Location
+</Button>
+
+<h2 class="h3">
+  {title}
+</h2>
+
+<div class="h4">
```
mukhtyar, line 34: use `p` element

src/routes/tools/extreme-precipitation/_ExploreData.svelte
```diff
@@ -0,0 +1,253 @@
...
+      indicatorLabel="{indicatorLabel}"
+      durationLabel="{durationLabel}"
+      scenarioLabel="{$scenario.labelLong}"
+      thresholdLabel="{thresholdLabel}"
+      intervalsLabel="{intervalsLabel}"
+      loadLocation="{loadLocation}"
+    />
+  </div>
+
+  <div slot="stats">
```
mukhtyar, line 196: Stats is not yet implemented. Will need to brainstorm what stats to show for the Intensity chart.

src/routes/tools/extreme-precipitation/_store.js
```diff
@@ -0,0 +1,101 @@
...
+    {
+      name: "intensity",
+      getter: ($s) => $s.intensity,
+    },
+    {
+      name: "events",
+      getter: ($s) => $s.events,
+    },
+    {
+      name: "frequency",
```
mukhtyar, line 67: Check why totals for frequency and duration are different from existing tool.

mukhtyar (commented):
A quick review of the current state of the tool. Need to work on stats and figure out the discrepancy between the totals from existing tool.

src/components/tools/Charts/DotAndWhisker/DotAndWhisker.svelte
```diff
@@ -31,16 +31,16 @@
...
           on:mouseover="{(e) => dispatch('mousemove', { e, props: d })}"
```
mukhtyar, line 35: Use the new mouseover actions once the changes from main are merged in.

content/tools/data.js
```diff
@@ -101,6 +101,27 @@ const tools = [
...
+        slug: '0887cba6-695c-403a-a757-674075799ea5',
+        logo: 'u_colorado.png',
+      },
+      {
+        slug: 'b8bee9f9-31c1-40d9-89a3-2dad814ba7da',
+        logo: 'gif_80x80.png'
+      }
+    ],
+    related: [
+      'extreme-weather','extreme-precipitation',
```
clhenrick: Should "extreme-precipitation" be here? Isn't that the same name as this tool?

src/components/tools/Charts/DotAndWhisker/Chart.svelte
```diff
@@ -89,33 +80,63 @@
...
           color: series.color,
           visible: true,
         };
       })
     );
-    console.log($legendItems);
     setContext("Legend", legendItems);
   }
+
+  function createTooltip(d) {
```
clhenrick: Since the ToolTip component is local to this Chart component only, why not pass in the props to the component and render the HTML there using Svelte's template syntax vs. doing string concatenation here which is more difficult to read / follow?

src/routes/tools/extreme-precipitation/_ChartTitle.svelte
```diff
@@ -0,0 +1,51 @@
+<script>
```
clhenrick, line 1: Wondering if we should make these Title components a Partial that have slots for the h2 and p elements. That way we wouldn't need to be rewriting the Change Location Button each time we need it.

src/routes/tools/extreme-precipitation/_store.js
```diff
@@ -0,0 +1,101 @@
...
+
+export const indicatorList = CLIMATE_INDICATORS;
+
+export const indicatorStore = (() => {
+  const store = writable(DEFAULT_CLIMATE_INDICATOR);
+  const { set, subscribe } = store;
+  return {
+    set,
+    subscribe,
+    get indicator() {
```
clhenrick: I'm curious if using `makeCustomWritableStore` does not work for this use case?

src/routes/tools/extreme-precipitation/_store.js
```diff
@@ -0,0 +1,101 @@
...
+  name: "climvarStore",
+  getters: [
+    {
+      name: "climvar",
+      getter: ($s) => climvars.find((d) => d.id === $s),
+    },
+  ],
+});
+
+export const indicatorList = CLIMATE_INDICATORS;
```
clhenrick, line 30: Would it make sense to make `indicatorList` a readable store if it's never updated?

clhenrick (commented):
@mukhtyar seems like a good start, left a couple comments / suggestions but just minor things.

I'm wondering if we should make the marks in this chart slightly bigger? Currently they seem like they could be a little difficult to read, especially over the white background:

Screen Shot 2021-12-22 at 10 36 23 AM

Also I'm seeing this issue again with the "About" section getting pushed down by the settings panel. I'm wondering if we might need to refactor the dashboard layout to prevent this some how.

Screen Shot 2021-12-22 at 10 37 32 AM

src/routes/tools/extreme-precipitation/_store.js
mukhtyar: I've left `indicatorList` a constant, to match the other lists used to populate settings e.g. `climvars`, `scenarios`, etc.

src/routes/tools/extreme-precipitation/_store.js
mukhtyar: Forgot about that one. Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/163/commits/fa01f01981bcfa571605ac8778a0911054ea9b3f

src/components/tools/Charts/DotAndWhisker/Chart.svelte
mukhtyar: Good idea. Implemented in https://github.com/berkeley-gif/caladapt-website-2021/pull/163/commits/ddcd93d89503c24014abce04eaa7ce0370831030

content/tools/data.js
mukhtyar: Removed it in https://github.com/berkeley-gif/caladapt-website-2021/pull/163/commits/c1124077a07acd5fd1673d143cca56a4510f3bdd

src/routes/tools/extreme-precipitation/_ChartTitle.svelte
mukhtyar: The partial might work for some tools. But others e.g. Wildfire, Snowpack have more complex Title components with multiple `p` elements and notifications.

mukhtyar:
> @mukhtyar seems like a good start, left a couple comments / suggestions but just minor things.
> 
> I'm wondering if we should make the marks in this chart slightly bigger? Currently they seem like they could be a little difficult to read, especially over the white background:
> 
> Screen Shot 2021-12-22 at 10 36 23 AM
> 
> Also I'm seeing this issue again with the "About" section getting pushed down by the settings panel. I'm wondering if we might need to refactor the dashboard layout to prevent this some how.
> 
> Screen Shot 2021-12-22 at 10 37 32 AM

@clhenrick The space issue with the About section should be resolved with the addition of the Stats panel. For the Timing chart, I have increased the height of the chart so the marks will appear bigger.

src/routes/tools/extreme-precipitation/_ChartTitle.svelte
clhenrick: True, though you could still wrap multiple `<p>` elements in a `div` to slot them. It brings up a good point though which is that standardizing these components is difficult!

src/components/tools/Charts/DotAndWhisker/Chart.svelte
```diff
@@ -18,67 +22,54 @@
...
-    const upper = Math.round(d.upperci * 100) / 100;
-    tooltip += `<span>Return level: ${val} ${yAxis.units}</span>`;
-    if (lower === 0 || upper === 0) {
-      tooltip += "<span>Insufficient observations to calculate CI</span>";
-    } else {
-      tooltip += `<span>95% CI: ${lower} – ${upper} ${yAxis.units}</span>`;
-    }
-    return tooltip;
-  }
+  $: if (!Array.isArray(data) || !data.length) {
```
clhenrick, line 55: I think you could use `isEmptyData(data)` here right?

src/components/tools/Charts/DotAndWhisker/Chart.svelte
```diff
@@ -89,33 +80,43 @@
       .align(0);
 
     // Update Y Domain
-    ymin = min(data, (d) => d.lowerci);
-    ymax = max(data, (d) => d.upperci);
-    if (yAxis.baseValue === 0) {
-      ymin = yAxis.baseValue;
+    if (typeof yAxis.domainMin === "number" && !isNaN(yAxis.domainMin)) {
```
clhenrick, line 83: might be good to extract this logic for testing a number into a helper function in `helpers/utilities` since we use it frequently within the tools.

src/routes/tools/extreme-precipitation/_store.js
```diff
@@ -0,0 +1,102 @@
...
+    getters: [
+      {
+        name: "indicator",
+        getter: ($s) => indicatorList.find((d) => d.id === $s),
+      },
+    ],
+  }
+);
+
+export const thresholdStore = writable(null);
```
clhenrick, line 45: just FYI it can still be useful to use `makeCustomWritableStore` for these because you can add the `name` getter which is useful for logging store values when they change.

clhenrick (approved):
Just a few comments feel free to merge and open a new PR for the stats panels or other updates as this one is quite large already.

PR #165, opened 2021-12-21
upgrade deps: svelte, svelte-loader, webpack

clhenrick:
Seems good to upgrade Svelte to the latest version but we should check to see if it breaks anything.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,7 +65,7 @@
         "svelte-inview": "^2.0.1",
         "svelte-loader": "2.13.6",
         "svelte-preprocess": "^4.6.1",
-        "webpack": "^4.7.0",
+        "webpack": "^4.46.0",
         "webpack-bundle-analyzer": "^4.5.0",
         "webpack-modules": "^1.0.0",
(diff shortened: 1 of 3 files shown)


clhenrick:
@mukhtyar I’m noticing there’s a lot of a11y warnings being outputted in the terminal, seems mostly from hover events on chart SVG elements. It would be good if we could suppress them so that there’s not tons of output in the console each time a (re)build happens

mukhtyar (approved):

mukhtyar:
@clhenrick created a [new PR](https://github.com/berkeley-gif/caladapt-website-2021/pull/166) to fix the a11y warnings

clhenrick:
@mukhtyar did you take some time to poke around the tools and make sure nothing broke with any of them? I haven't yet but will today, would be good to double check.

clhenrick:
@mukhtyar I also ended up upgrading `svelte-loader` and `webpack`; hopefully this helps make builds a little faster.

clhenrick:
Wait until #166 is merged before merging this PR.

PR #166, opened 2021-12-21
Feature fix chart a11y warnings

mukhtyar:


src/components/tools/Charts/Shared/events.js
```diff
@@ -0,0 +1,23 @@
+export const mouseout = (node, callback) => {
```
clhenrick: It might be better to name these `mouseoutBlur` and `mouseoverFocus` as they're handling both events.

src/components/tools/Charts/Histogram/Column.svelte
```diff
@@ -47,8 +48,8 @@
       fill-opacity="{0.2}"
       stroke="{stroke}"
       stroke-width="{strokeWidth}"
-      on:mouseout="{() => dispatch('mouseout')}"
-      on:mouseover="{(e) => dispatch('mousemove', { e, props: d })}"
+      use:mouseout="{() => dispatch('mouseout')}"
```
clhenrick: It's worth noting that although Svelte is complaining about a11y here, these elements won't even be reachable by keyboard or AT unless they were given extra attributes such as tabindex, at least as far as I know.

src/components/icons/App.svelte
```diff
@@ -23,6 +23,7 @@
 <svg
   on:click
```
clhenrick: Do we need any of these event handlers for the tool/climate icons? I think the only place they were formerly interactive was on the home page. If they're to be used interactively in the future then wrapping them in a `<button>` element would be a more appropriate way to handle these event listeners.

clhenrick: In fact, we should probably remove most of the aria related & title props. The `<svg>` element should have an `alt` attribute that is an empty string (`""`) if no alt text is given or a description of what the icon's purpose is otherwise (e.g. `alt= "Filter climate tools by sun category"`). 

Using `aria-labelledby` in combination with the SVG `<title>` element is a better alternative to using `alt` text for inlined SVG, however because all of these icon components use the same `id` for their title, that will mess things up. 

To fix the a11y my recommendation would be to 

1. add a unique id for each title, e.g. `icon-title-rainfall`
2. use `aria-labelledby="unique-icon-title-id"` on the SVG element
3. add `role="img"` on the SVG element

The contents of the SVG title element could be dynamic via a prop on the component with the fallback being "<Icon Name> Icon".

clhenrick: This is a good article on accessible SVG btw: https://css-tricks.com/accessible-svgs

clhenrick (commented):
Thanks @mukhtyar. It would be good to clean up the a11y on the icon components, see my comments below.

src/components/tools/Charts/Shared/events.js
mukhtyar: Good idea. Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/166/commits/96d50f150c835ff45c6d317b91ed140d322ea1f3

src/components/icons/App.svelte
mukhtyar: Thanks for the recommendations! Let me know if the fixes in https://github.com/berkeley-gif/caladapt-website-2021/pull/166/commits/8c97c84274795025eeb80d24c5d57174952a4a21 are correct.

src/components/icons/App.svelte
```diff
@@ -1,42 +1,26 @@
...
   export let title = undefined;
   export let style = undefined;
   export let dimension = 60;
 
-  $: ariaLabel = $$props["aria-label"];
-  $: ariaLabelledBy = $$props["aria-labelledby"];
-  $: labelled = ariaLabel || ariaLabelledBy || title;
+  let id = "icon-title-app";
   $: attributes = {
+    id,
```
clhenrick: @mukhtyar it seems like the `id` is being applied to both the svg and title elements, which means it's not unique. More info here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

src/components/icons/App.svelte
```diff
@@ -1,42 +1,26 @@
...
-  class="{className}"
   preserveAspectRatio="xMidYMid meet"
-  style="{style}"
-  id="{id}"
   viewBox="0 0 32 32"
-  aria-labelledby="iconTitle"
   {...attributes}
 >
-  <title id="iconTitle">All Apps</title>
+  <title id="{id}">{title ? title : "Icon"}</title>
```
clhenrick: "Icon" isn't a meaningful title, especially if multiple SVG elements share the same title. Maybe make the default "App Icon", "Moon Icon", etc.

src/components/icons/App.svelte
clhenrick: No problem. I left a couple comments on the App Icon file.

src/components/icons/App.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/166/commits/9ff2407e77162347d964395260101aa905d46fd5

src/components/icons/App.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/166/commits/9ff2407e77162347d964395260101aa905d46fd5

src/components/icons/App.svelte
clhenrick: @mukhtyar I'm realizing now the a11y fix I mentioned previously is only useful for when the SVG is informative for a user. However, we are only using these icons in a decorative manner in the UI. AFAIK no where are they clickable, meaning we're not using them wrapped in a button with no visual text. For that reason I think it would make sense to either: 

1. Have the title text be a prop that defaults to `undefined`. If the value is truthy, then we would render the `<title>` within the SVG with the title text as well as add the `aria-labelled-by="title-id"` and `role="img"`. If the value of the title text prop is falsey, then we would add `role="presentation"` and `aria-hidden=true` to the SVG element as well as not render the `<title>`.

2. Strip out all of the following SVG id, aria-labelled-by, `<title>` and hardcode the `role="presentation"` and `aria-hidden=true` attributes to the SVG making the icons only decorative.

[Here's an article that explains more](https://scottvinkle.me/blogs/work/svg-and-emoji-icon-accessibility)

I apologize for the back and forth on this and I'm happy to make the changes if you'd prefer not to.

mukhtyar: @clhenrick No worries. I can make the updates. You have enough on your plate with the SLR tool.

clhenrick (commented):
LGTM!

PR #170, opened 2022-01-06
SLR CIS: delay fetching of raster metadata

clhenrick:
Instead of fetching raster metadata each time the map's `moveend` event fires, add a debounce to the event handler so that it will fire after sequential map `moveend` events, for example if a user is repeatedly zooming and/or panning.


clhenrick:
[SLR CIS: add delay for data fetching when user zooms or pans map](https://trello.com/c/AKRNyrtp/152-slr-cis-add-delay-for-data-fetching-when-user-zooms-or-pans-map)

PR #179, opened 2022-01-22
Bug fix load bookmark params

mukhtyar:


src/routes/tools/_common/constants.js
```diff
@@ -100,10 +100,10 @@ export const OBSERVED = [
 export const OBSERVED_FILTER_YEAR = 2006;
 
 export const INITIAL_CONFIG = {
-  boundaryId: "locagrid",
-  scenarioId: "rcp45",
-  climvarId: "tasmax",
-  modelIds: ["HadGEM2-ES", "CNRM-CM5", "CanESM2", "MIROC5"],
+  boundary: "locagrid",
```
mukhtyar, line 103: Changed these property names to match what's used for the share link. This makes it easier to overwrite any properties from the share link.

src/routes/tools/annual-averages/index.svelte
```diff
@@ -4,8 +4,8 @@
   // the data we need to render the page. It only runs once
   // during export.
   import resourcesList from "../../../../content/resources/data";
-  import { INITIAL_CONFIG } from "../_common/constants";
-  export async function preload({ query }) {
+
+  export async function preload(page) {
```
mukhtyar: Remove `page`.

src/routes/tools/_common/helpers.js
```diff
@@ -234,3 +234,26 @@ export const convertAnnualRateToSum = ({ date, value }) => {
...
 };
+
+/**
+ * Create initial configuration
+ * @param {object} defaultParams - default params for tool
+ * @param {object} urlParams - params generated from url query string by sapper
+ * @return {object}
+ */
+// Helper function to create an object with initial settings for a Cal-Adapt tool
+export const getInitialConfig = (defaultParams = {}, urlParams = {}) => {
```
clhenrick: I might reverse these function params so that `urlParams` is first and `defaultParams` is second. Then you could have `defaultParams` default to the `INITIAL_CONFIG` common const and might not need to import it in the tool. That would make using this function a little simpler.

src/routes/tools/_common/helpers.js
```diff
@@ -234,3 +234,26 @@ export const convertAnnualRateToSum = ({ date, value }) => {
...
+  if (Object.keys(urlParams).length === 0) {
+    return defaultParams;
+  }
+
+  const { models, lat, lng, ...rest } = urlParams;
+
+  return {
+    ...defaultParams,
+    ...rest,
+    models: models.split(","),
```
clhenrick: what if `models`, `lat`, or `lng` aren't in the `urlParams`? They will still override the values from the `defaultParams` but set them to `undefined` or `NaN` (I think).

src/routes/tools/annual-averages/index.svelte
```diff
@@ -67,10 +46,13 @@
   import { onMount } from "svelte";
   import { Loading } from "carbon-components-svelte";
   import { inview } from "svelte-inview/dist/";
+  import { stores } from "@sapper/app";
```
clhenrick: Because we're typically importing so many stores, I might alias this as `sapperStores` so that it's a little more clear in the code what it's referring to. e.g.

```js
import { stores as sapperStores } from "@sapper/app";
```

clhenrick (commented):
Makes sense to extract the logic for reconciling the default config with the query params config values. Just left a couple minor comments.

src/routes/tools/_common/helpers.js
mukhtyar: @clhenrick yes that's a tricky one. I added a check to see if the prop exists before adding it in https://github.com/berkeley-gif/caladapt-website-2021/pull/179/commits/1f423dce93618c7cff447a826c99087467555364

src/routes/tools/_common/helpers.js
mukhtyar: @clhenrick I tried that first and then left it here because sometimes we want to start with a different `INITIAL_CONFIG`. See the update for the Degree Days tool in https://github.com/berkeley-gif/caladapt-website-2021/pull/179/commits/0fe226317029d5b3f5c9620779f1ab285df22111. But will think about this a bit more.

src/routes/tools/annual-averages/index.svelte
mukhtyar: good idea. fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/179/commits/11a328605b27810fc05273d28ed06df7fa1a6e40

src/routes/tools/_common/helpers.js
clhenrick: But you could still override it with the function params in the opposite order right? Not following why that doesn't work.

mukhtyar: I see what you're saying. Never mind. Switched the param order in https://github.com/berkeley-gif/caladapt-website-2021/pull/179/commits/0529d0355f4066fd8772dab449ab63b3da48bc04.

src/components/tools/Settings/SelectDayOfYear.svelte
```diff
@@ -11,9 +11,10 @@
   $: placeholder = timeFormat(formatStr)(value);
 
   const dispatch = createEventDispatcher();
+  const currentYear = +new Date().getUTCFullYear();
```
clhenrick, line 14: Wondering if this might be a global constant we should make available for the rest of the website / app?

Also does this component affect the bookmark bug?

src/routes/tools/_common/constants.js
clhenrick: ~~Will this break the other tools in any way? Do we need to change their names in other tools to prevent any bugs?~~

src/routes/tools/_common/helpers.js
```diff
@@ -234,3 +235,28 @@ export const convertAnnualRateToSum = ({ date, value }) => {
     return value * 365;
   }
 };
+
+/**
+ * Create initial configuration
+ * @param {object} defaultParams - default params for tool
```
clhenrick: maybe switch the order of these `@param`s in the doc string to reflect their order as function args

src/routes/tools/_common/helpers.js
```diff
@@ -234,3 +235,28 @@ export const convertAnnualRateToSum = ({ date, value }) => {
...
+  if (Object.keys(urlParams).length === 0) {
+    return defaultParams;
+  }
+
+  const { models, months, ...rest } = urlParams;
+
+  return {
+    ...defaultParams,
+    ...rest,
+    ...(models && { models: models.split(",") }),
```
clhenrick: What might make setting these optional properties more resilient is if instead of `models.split(",")` we passed `models` to a helper function that validated the value of `models` before splitting it into an array. For example:

```js
function validateModels(value) {
  return typeof value === "string" && value.split(",").length
}
```

Then we could do:

```
...(validateModels(models) && { models: models.split(",") }),
```

Something similar could be done for months, and depending on how strict we want to be on validating the values we could even make sure they match expected values like model names or month abbreviations, etc. 

But at the very least I think testing that the value is a string and can be split into an array would be a good way to prevent potential corrupt data for these.

src/routes/tools/_common/constants.js
clhenrick: Nevermind I see that you've renamed things in the other tools, thanks!

clhenrick (approved):
Wow it's awesome to see how much duplicate code this change removes. 

I haven't yet tried using all the tools to see if anything is broken, I assume you have though and I could take a look at some specific tools more in-depth if you'd like, let me know. 

One way of having reassurance that when we make changes like this across all tools that no individual breaks would be having some end to end tests, but they can be a bit of work to set up. We might want to discuss it later.

src/components/tools/Settings/SelectDayOfYear.svelte
mukhtyar: The Day of Year Number used within the Extreme Weather tool was using different years while parsing & formating. I noticed this while testing the bookmark and hence made the change here.

src/routes/tools/_common/helpers.js
mukhtyar: @clhenrick I added validation functions for `models` and `months`. Let me know what you think.

```
function validateModels(value) {
  if (typeof value !== "string" || !value.split(",").length) {
    return false;
  }
  const models = PRIORITY_10_MODELS.map(({ id }) => id);
  return value.split(",").every((d) => models.includes(d));
}

function validateMonths(value) {
  if (typeof value !== "string" || !value.split(",").length) {
    return false;
  }
  const range = extent(MONTHS_LIST, (d) => d.id);
  return value.split(",").every((d) => +d >= range[0] && +d <= range[1]);
}
```

src/routes/tools/_common/helpers.js
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/179/commits/b870429316f0dcfe4f28a04b44892ba9ec3519e9

src/routes/tools/_common/helpers.js
clhenrick: That's definitely a more robust way of checking those properties. The only improvement I would try to make is to reduce the number of times `String.prototype.split` is used. Currently it will be used 2 times in each validation method and then again when actually setting the property. 

I might try doing the following:

1. test if the `value` is a string and has a truthy `length` property
2. make a local variable for the array from `value.split(",")`
3. do the validation test over each item in the array
4. if it's valid, add the property to the config object
5. if not then don't add the property to the config object

That way you would only need to use `.split(",")` once, but it might mean you have to pass the config object into the validation helper function instead of using the optional object property syntax with the spread operator.

There's probably another way to do it too, let me know if you think of it, no pressure to do any of this though either.

src/routes/tools/_common/helpers.js
```diff
@@ -234,3 +237,54 @@ export const convertAnnualRateToSum = ({ date, value }) => {
...
+  if (Object.keys(urlParams).length === 0) {
+    return defaultParams;
+  }
+
+  const { models, months, ...rest } = urlParams;
+
+  return {
+    ...defaultParams,
+    ...rest,
+    ...(validateModels(models) && { models: validateModels(models) }),
```
clhenrick, line 287: @mukhtyar just FYI you could also stash the value of `validateModels` and `validateMonths` in local variables instead of calling the function twice.

PR #180, opened 2022-01-25
Bug fix location names not updating

mukhtyar:
This PR fixes the issue of Chart Titles not updating in multiple tools.


clhenrick (commented):
@mukhtyar thanks for fixing these. I'm wondering if we need to use `await tick()` somewhere or what the root cause could be. I think it's likely an async process that isn't completing before the title is set as the `location` store is updated async.

mukhtyar:
@clhenrick The location name updates in these 3 tools worked before. I was wondering if updating the svelte version has something to do with this issue. I'll experiment with `await tick()` in the `streamflow` tool.

PR #181, opened 2022-01-26
SLR CIS beta preview updates

clhenrick:
- [x] default to dark style basemap
- [x] add draft about content
- [x] add resources
- [x] update missing data message


clhenrick:
[SLR CIS: beta updates prior to CEC review](https://trello.com/c/l8BaRe64/157-slr-cis-beta-updates-prior-to-cec-review)

src/routes/tools/slr-coastal-inundation/index.svelte
```diff
@@ -232,7 +238,7 @@
       on:datasetLoaded="{(e) => datasetStore.set(e.detail)}"
     >
       <div slot="description">
-        <p>To do...</p>
+        {@html aboutContent}
```
mukhtyar, line 241: This feels like a better solution than having all the text here.

mukhtyar (approved):

src/routes/tools/slr-coastal-inundation/index.svelte
clhenrick: Thanks @mukhtyar I wonder if we might be able to leverage it for other pieces of content such as in the "learn more" content modals?

PR #182, opened 2022-01-26
Bug fix: Loading the tool with the Share Link errors for some locations

mukhtyar:
The PR switches the order of how the initial location is set in the  `initApp` function. Instead of requesting nearest Mapbox feature first, the intersecting polygon is requested from Cal-Adapt API first. The nearest Mapbox feature is only requested for `locagrid` to update the feature title.


mukhtyar:
@clhenrick you can test this out on the dev site. The following url should now load the tool without breaking it:

https://dev.cal-adapt.org/tools/degree-days?climvar=tasmax&frequency=M&indicator=hdd&scenario=rcp85&threshold=67&models=CNRM-CM5%2CCanESM2&months=6%2C7&lng=-120.3614&lat=39.5731&boundary=place

clhenrick (commented):
@mukhtyar I would suggest having a fallback in case either the call to `getFeature` or `reverseGeocode` fails so that the app doesn't break. You might even wrap this code in a helper function as it's duplicated in all of the tools.

This is sort of a naive implementation but maybe something along these lines:

```js
async function maybeSetPlaceName(lat, lng, boundary) {
  let loc = FALLBACK_LOCATION;
  
  try {
    loc = await getFeature({ center: [+lng, +lat] }, boundary);
  } catch (error) {
    console.warn(error);
  }

  if (boundary === "locagrid") {
    let placeName = DEFAULT_PLACE_NAME;
    
    try {
      const result = await reverseGeocode(`${lng}, ${lat}`);
      
      if (result && result.features && result.features.length) {
        placeName = result.features[0].place_name;
      }
    } catch (error) {
      console.warn(error);
    } finally {
      loc.title = getTitle(loc, boundary, placeName);
    }
  }

  return loc;
}
```

then in the `initApp` you could do

```js
const loc = maybeSetPlaceName(lat, lng, boundary);
locationStore.updateLocation();
```

clhenrick (approved):
LGTM!

PR #185, opened 2022-01-27
dev chore: tools use markdown for "about" section

clhenrick:
Where possible, update the migrated tools About section to use content from a Markdown file and render it with Svelte `@html` so that it reduces the size of the JS bundle.


src/routes/tools/extreme-weather/index.json.js
```diff
@@ -0,0 +1,22 @@
+import { getToolContent } from "../_common/server-utils";
+
+export function get(_req, res) {
+  try {
+    const {
+      description,
+    } = require("/content/tools/extreme-weather/description");
```
clhenrick, line 7: Note: for content that is better suited to live in `.js` or `.json` files we can just use `require` instead of `fs.readFileSync`. Although I should probably update the `getToolContent` function to handle that...

src/routes/tools/wildfire/index.svelte
```diff
@@ -30,11 +30,16 @@
       ["get-started", "faqs"].includes(d.slug)
     );
 
+    const { aboutContent } = await (
+      await this.fetch("tools/annual-averages.json")
```
clhenrick: should be `wildfire.json`

src/routes/tools/extreme-weather/index.json.js
clhenrick: It appears that NodeJS / Webpack / Sapper won't let you use `require` from an imported function so leaving as is.

mukhtyar (approved):
Looks good!

PR #188, opened 2022-02-03
SLR CIS: replace change location widget with map search box

clhenrick:
Updates the SLR CIS tool to use a search box within the map area instead of the Set Location widget that's common with other tools, as this tool doesn't perform any analysis and thus does not require a location boundary.

### DIFF
--- a/src/components/tools/Map/Map.svelte
+++ b/src/components/tools/Map/Map.svelte
@@ -45,6 +45,10 @@
     }
   }
 
+  export function flyTo(center, zoom, options = {}) {
+    map.flyTo({ center, zoom, ...options });
+  }
+
   export function resize() {
     map && map.resize();
   }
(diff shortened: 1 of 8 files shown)


clhenrick:
[SLR CIS: replace change location with map search box](https://trello.com/c/NWvN3qko/163-slr-cis-replace-change-location-with-map-search-box)

clhenrick:
@mukhtyar the way the search suggestions are shown with the new Map/Search component are a bit more natural than how they are shown in the ChangeLocation widget. By debouncing the input event we can still give the user a natural autosuggest UX while not making too many requests to the geocoding APIs. Let me know if you have any thoughts on it, thanks!

mukhtyar (approved):
The debounced map search input works really well! Definitely a better user experience. And not having the Location widget also helps in clarifying to the user how to navigate the tool.

PR #191, opened 2022-02-11
Bug fix stats support keyboardentry

mukhtyar:
Currently typing in a year in the Change Stats dialog (instead of selecting year from the combobox dropdown menu) for a Stats box returns `N/A`. This PR includes the foll:
- simplifies getting index of a value in the item list
- shows an invalid state if the `start` & `end` year are missing for custom period
- adds support for entering a year using the keyboard
- filters list as user types in a year
- adds comments

### DIFF
--- a/src/components/tools/Stats/StatOptions.svelte
+++ b/src/components/tools/Stats/StatOptions.svelte
@@ -55,8 +55,8 @@
    * the combobox selected index if the year matches an item in the combobox dropdown menu
    **/
   function updateIndex(e) {
-    const { key, keyCode, target } = e;
-    if (["Enter", "Tab"].includes(key) || [13, 9].includes(keyCode)) {
+    const { key, target } = e;
+    if (["Enter", "Tab"].includes(key)) {
       const { value, id } = target;
       switch (id) {
         case "years-select-start":
--- a/src/components/tools/Stats/StatOptions.svelte
+++ b/src/components/tools/Stats/StatOptions.svelte
@@ -46,6 +46,9 @@
   // If user confirms changes, dispatch change event with selected groupd and period
   // If user has selected a custom period, create a new period object
   function update() {
+    if (startYear_selectedIndex < 0 || endYear_selectedIndex < 0) {
+      return;
+    }
     group = groupList.find(({ id }) => id === selectedGroupId);
     if (selectedPeriodId === "custom") {
       const start = getItem(startYear_selectedIndex);
@@ -102,12 +105,16 @@
         titleText="Start"
         placeholder="Select start year"
         items="{items}"
+        invalid="{startYear_selectedIndex < 0}"
+        invalidText="Select Start Year"
       />
       <ComboBox
         bind:selectedIndex="{endYear_selectedIndex}"
         titleText="End"
         placeholder="Select end year"
         items="{filteredItems}"
+        invalid="{endYear_selectedIndex < 0}"
+        invalidText="Select End Year"
       />
     </div>
   {/if}
--- a/src/components/tools/Stats/StatOptions.svelte
+++ b/src/components/tools/Stats/StatOptions.svelte
@@ -26,9 +26,7 @@
   let endYear_selectedIndex = -1;
   let filteredItems;
 
-  const getItem = (i) => (items[i] ? +items[i].text : "N/A");
-  const getFilteredItem = (i) =>
-    filteredItems[i] ? +filteredItems[i].text : "N/A";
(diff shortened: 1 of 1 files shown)


mukhtyar:
[Bug: Keyboard input of custom year in the Stats box results in N/A](https://trello.com/c/evPxXeik/174-bug-keyboard-input-of-custom-year-in-the-stats-box-results-in-n-a)

src/components/tools/Stats/StatOptions.svelte
```diff
@@ -39,17 +41,56 @@
...
+    return item.text.includes(value);
+  }
+
+  /**
+   * When user presses the Enter or Tab key in combobox after typing in a year, update
+   * the combobox selected index if the year matches an item in the combobox dropdown menu
+   **/
+  function updateIndex(e) {
+    const { key, keyCode, target } = e;
+    if (["Enter", "Tab"].includes(key) || [13, 9].includes(keyCode)) {
```
clhenrick: I think with modern browsers we can get away with only using `key` and don't need to also check for `keyCode`. Using `keyCode` can be more error prone from my understanding.

src/components/tools/Stats/StatOptions.svelte
```diff
@@ -39,17 +41,56 @@
...
+      const { value, id } = target;
+      switch (id) {
+        case "years-select-start":
+          startYear_selectedIndex = items.findIndex((d) => d.text === value);
```
clhenrick: I think part of the bug I'm seeing has to do with how `startYear_selectedIndex` is being set here.

clhenrick (commented):
@mukhtyar I'm still seeing a bug where if I type in a partial year like `203` and then use the up and down arrow keys to select an item from the list, when I hit `enter` it will set the input to value that isn't the one I selected. Here's a gif showing the issue:

Feb-14-2022 10-16-42

src/components/tools/Stats/StatOptions.svelte
clhenrick: If `value` is a string like "203" then this won't work as expected.

src/components/tools/Stats/StatOptions.svelte
```diff
@@ -39,17 +41,56 @@
...
+            (d) => d.text === value
+          );
+          return;
+        default:
+          return;
+      }
+    }
+  }
+
   $: yearsList = range(dateRange[0], dateRange[1] + 1, 1);
```
clhenrick: @mukhtyar these `$` variables could be moved above the `function` declarations so that they're easier to discover when first reading the code.

src/components/tools/Stats/StatOptions.svelte
mukhtyar: Got it. Fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/191/commits/ffbc1fc9ff584715f7aa9a2759ffeb9e2309e973

src/components/tools/Stats/StatOptions.svelte
mukhtyar: @clhenrick I updated the code to set index to highlighted item index in https://github.com/berkeley-gif/caladapt-website-2021/pull/191/commits/9be8ac2d7337a2059b30e05e7b7d4076b6c2f0de. It's possible to get the right index, but ComboBox component still overrides it.

src/components/tools/Stats/StatOptions.svelte
mukhtyar: good idea. fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/191/commits/5a00c0d20c0db4938f704824e06a6611cba1425a

src/components/tools/Stats/StatOptions.svelte
clhenrick: That's really strange. So even after you find the correct index using the highlighted class name it still is overridden?

mukhtyar: Yes. Looking at the code, it look like the component also tracks highlighted id internally. `shouldFilterItem` runs after each character is entered and at this point there are 2 lists for the combobox -the  original with all values and a temporary filtered list with matching values. The index of the value highlighted (e.g. 2032) might be 26 in original list but 2 in the filtered list. So after I set the `selectedIndex` to 26 (corresponding to 2032), some lifecycle function in the component probably sets it to 2 (corresponding to 2008).

clhenrick:
superseded by #194

PR #194, opened 2022-02-15
bug fix: StatOptions component keyboard event

clhenrick:
@mukhtyar this seems to fix the bug with the Stats keyboard event listener, let me know if it works for you.


mukhtyar:
@clhenrick perfect! The `await tick()` does the trick. Thanks.

clhenrick:
@mukhtyar so should we merge this one? If so I will add a code comment before merging.

mukhtyar (approved):
Looks good. We can merge this in.

PR #197, opened 2022-02-16
Feature extremeprecip add aggregation

mukhtyar:
This PR adds a Select Aggregate Function Setting & some notes to the Extreme Precipitation tool

### DIFF
--- a/content/tools/extreme-precipitation/about.md
+++ b/content/tools/extreme-precipitation/about.md
@@ -12,7 +12,7 @@
   override these defaults by selecting a new “event duration” (number of
   days over which precipitation accumulates), or by selecting a
   different “threshold“ value that corresponds to either the 90th, 95th
-  or 99th percentiles. The tool then displays the extreme events that
+  or 99th percentiles. The tool displays the extreme events that
   exceed the threshold in different ways. The Frequency chart shows the
   estimated intensity of precipitation events (Return Level) for a
   selected period (Return Period) and how it changes over the historical
@@ -95,6 +95,11 @@
   encourage users to ensure the empirical fit of the applied
   distribution is acceptable to their end use before using estimates
   produced from this tool for planning purposes.
+</p>
+
+<p>
+  <strong>Note</strong>: Based on feedback from our users, we have updated the tool to use the 
+  average of precipitation values from all intersecting grid cells for a polygon boundary before calculating metrics for the climate indicators. An earlier version of this tool used the maximum values from all intersecting grid cells which resulted in very high frequency estimates. The return levels estimates using this methodology are more closely align with precipitation frequency estimates provided by NOAA.
 </p>
 
 <p class="h3">References</p>
--- a/content/tools/extreme-precipitation/about.md
+++ b/content/tools/extreme-precipitation/about.md
@@ -98,8 +98,8 @@
 </p>
 
 <p>
-  <strong>Note</strong>: Based on feedback from our users, we have updated the tool to use the 
-  average of precipitation values from all intersecting grid cells for a polygon boundary before calculating metrics for the climate indicators. An earlier version of this tool used the maximum values from all intersecting grid cells which resulted in very high frequency estimates. The return levels estimates using this methodology are more closely align with precipitation frequency estimates provided by NOAA.
+  <strong>Update (Feb 20, 2022)</strong>: Based on feedback from our users, we have updated the tool to use the 
+  average of precipitation values from all intersecting grid cells for a polygon boundary before calculating metrics for the climate indicators. An earlier version of this tool used the maximum values from all intersecting grid cells which resulted in very high frequency estimates. The return levels estimates using this methodology more closely align with precipitation frequency estimates provided by NOAA.
 </p>
 
 <p class="h3">References</p>
--- a/content/tools/extreme-precipitation/about.md
+++ b/content/tools/extreme-precipitation/about.md
@@ -99,7 +99,7 @@
 
 <p>
   <strong>Update (Feb 20, 2022)</strong>: Based on feedback from our users, we have updated the tool to use the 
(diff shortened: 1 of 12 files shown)


src/routes/tools/extreme-precipitation/index.svelte
```diff
@@ -34,7 +34,7 @@
       ["get-started", "faqs"].includes(d.slug)
     );
 
-    const { aboutContent, learnMoreContent } = await (
+    const { aboutContent, learnMoreContent, notificationText } = await (
```
mukhtyar, line 37: remove unused var `notificationText`

content/tools/extreme-precipitation/about.md
```diff
@@ -97,6 +97,11 @@
   produced from this tool for planning purposes.
 </p>
 
+<p>
+  <strong>Update (Feb 20, 2022)</strong>: Based on feedback from our users, we have updated the tool to use the 
+  average of precipitation values from all intersecting grid cells for a polygon boundary before calculating metrics for the climate indicators. An earlier version of this tool used the maximum values from all intersecting grid cells which resulted in very high frequency estimates. The return levels estimates using this methodology more closely align with precipitation frequency estimates provided by NOAA.
```
mukhtyar: simplify text

mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/197/commits/fc8142df38d2a32d4407325cae48adde99e76aed

src/routes/tools/extreme-precipitation/index.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/197/commits/6edf6aa0bf7967e1d5d4ed67b7dc50bc4fd1aa84

src/routes/tools/extreme-precipitation/_constants.js
```diff
@@ -105,3 +115,7 @@ export const DEFAULT_INITIAL_CONFIG = {
 };
 
 export const HEATMAP_COLOR_SCALE = ["#19cdbb", "#368ce1", "#6d41ab"];
+
+export const NOTIFICATION_TEXT = `This tool now uses average precipitation values from all
```
clhenrick: This could live in a markdown file but not a big deal if it doesn't.

src/routes/tools/extreme-precipitation/index.svelte
```diff
@@ -152,7 +153,11 @@
   };
 
   $: $returnPeriodStore, updateIntensity();
-  $: $thresholdTypeStore, $location, $durationStore, updateThreshold();
+  $: $aggregateFnStore,
```
clhenrick, line 158: More of a tangential question, but I'm wondering if there's any benefit of making the function call reactive this way vs. just passing in the store values as the function params? In the SLR CIS tool I ended up doing the latter (I think) and it works.

clhenrick (commented):
@mukhtyar would it make sense to hide the "Select Aggregate Function" in the Settings Panel when the location boundary type is loca? It won't have an effect and could potentially confuse users otherwise, I think.

Also is "aggregate function" a term most users will understand? Just wondering if the wording could be different to be more user friendly, or not. Totally fine if not!

clhenrick:
@mukhtyar or perhaps instead of hiding the setting for aggregate function we should disable it when the location boundary type is loca grid cell, and make sure that in its "learn more" that this has no effect when boundary is "none".

src/routes/tools/extreme-precipitation/_constants.js
mukhtyar: moved to markdown in https://github.com/berkeley-gif/caladapt-website-2021/pull/197/commits/67f01172695dec6569e107f39994d987d24b5da9

mukhtyar:
@clhenrick good idea about disabling the `Select Spatial Aggregation` setting for LOCA. Implemented in https://github.com/berkeley-gif/caladapt-website-2021/pull/197/commits/eb84d0fb06cefb5c7c8b544083e6c4bf4d263ecd

src/routes/tools/extreme-precipitation/_SettingsPanel.svelte
```diff
@@ -27,16 +31,24 @@
...
+    periodInfo,
+    thresholdTypeInfo,
+    aggregationInfo,
+  } = learnMoreContent;
   const dispatch = createEventDispatcher();
-  const { location } = locationStore;
+  const { location, boundary } = locationStore;
   const { indicator } = indicatorStore;
 
+  $: isPolygonBoundary = $boundary.id === "locagrid" ? false : true;
```
clhenrick: @mukhtyar you can simplify this by just doing 

```js
$: isPolygonBoundary = $boundary.id !== "locagrid";
```

almost any time you're doing an equality comparison and want the result to literally be `true` or `false` it's a good sign you don't need a ternary expression.

clhenrick (commented):
Looks good, I'm wondering if we might want to add a note that spatial aggregation is not applied when the boundary type is a loca grid cell?

> Extreme Precipitation indicators for polygon boundaries (e.g. a watershed) are calculated after spatially aggregating precipitation values for all intersecting grid cells. The default spatial aggregation option is Mean. Selecting the Maximum option aggregates the maximum values from all intersecting grid cells.

src/routes/tools/extreme-precipitation/_SettingsPanel.svelte
clhenrick: I'm also thinking it might be good to have common tool constants for the boundary types, for example:

```js
// src/routes/tools/_common/constants.js
export const BOUNDARY_LOCA_GRID = "locagrid";
export const BOUNDARY_COUNTY = "county";
export const BOUNDARY_WATERSHED = "watershed";
// etc.
```

it might help with preventing misspelling the boundary types, but we don't have to do it right now.

mukhtyar:

PR #198, opened 2022-02-16
dev chore: expose feature flags to sapper server

clhenrick:
This enables feature flags from `featureFlags.json` to be used in server side code such as in Sapper / Svelte preload functions.  This is useful for things like hiding a new tool on production but enabling it on development or beta.

Noting that in the server side code feature flags are available under `process.cal_adapt_features`. For some reason I was running into problems when trying to add them to `process.env`, probably because it was messing with Node's built-in `process.env`.

~~As a bonus this PR also adds `serve` as a dev dependency so that running `start:export` doesn't download it every time.~~


mukhtyar (approved):
Looks good to me.

clhenrick:
Should probably update the readme to include a note about the feature flags.

PR #201, opened 2022-02-18
SLR CIS tool: clean up

clhenrick:
This PR removes dead code and cleans some things up in the SLR CIS tool.
- for cal-adapt api calls for `/rstores` `bbintersects` query param, use envelope instead of geojson 
- remove dep `@turf/bbox-polygon`
- remove unused Svelte stores
- make `MapLayerHandler` instances non-reactive

### DIFF
--- a/src/routes/tools/slr-coastal-inundation/_Map/Rasters.svelte
+++ b/src/routes/tools/slr-coastal-inundation/_Map/Rasters.svelte
@@ -19,8 +19,9 @@
     "raster-opacity": 0.5,
   };
 
-  const mapLayersProps = ({ id, checked, color, tileUrls }) => ({
+  const mapLayersProps = ({ id, checked, color, tileUrls, slugs }) => ({
     id,
+    slugs,
     tileUrls: tileUrls.map((url) => `${url}?style=${color}`),
     visibility: checked ? VISIBLE : NONE,
   });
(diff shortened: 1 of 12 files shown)


src/routes/tools/slr-coastal-inundation/_Map/Rasters.svelte
```diff
@@ -48,12 +49,11 @@
   });
 
   function addRasterLayers() {
-    rasterLayersProps.forEach(({ id, tileUrls, visibility }) => {
+    rasterLayersProps.forEach(({ tileUrls, visibility, slugs }) => {
       tileUrls.forEach((url, index) => {
         layerHandler.addMapLayer({
-          id,
+          id: slugs[index],
```
clhenrick, line 54: using a combination of a layer id and and index value to create unique MapBoxGLJS layer and source ids seems error prone; it's better to just use a unique id from the start, hence using the tile layer slug instead.

src/routes/tools/slr-coastal-inundation/_Map/TileIndexes.svelte
```diff
@@ -23,7 +23,7 @@
   });
 
   const mapLayersProps = ({ id, checked, color }) => ({
-    id,
+    id: `${id}-tile-index`,
```
clhenrick, line 23: Not really needed now that the raster layers are using unique ids, but it helps for debugging.

clhenrick (commented):
Fixed unique ids for map layers.

src/routes/tools/slr-coastal-inundation/_Map/Rasters.svelte
mukhtyar: good to know

mukhtyar (approved):
Looks good to me

PR #206, opened 2022-02-22
Migrate Extreme Precipitation Tool - QA #3 updates

mukhtyar:
This PR addresses the following issues:
- minor text updates in about and learn more sections
- adding a check and message to warn users about reduced certainty in estimates for some locations
- scaling the Intensity chart so it's easier to see confidence intervals

### DIFF
--- a/src/routes/tools/extreme-precipitation/_ExploreData.svelte
+++ b/src/routes/tools/extreme-precipitation/_ExploreData.svelte
@@ -265,7 +265,7 @@
       units="{$climvar.units.imperial}"
       label="{indicatorLabel}"
       dataSource="{$titles.join(', ')}"
-      niceMax="{$indicator.id === 'indicator' ? 5 : 10}"
+      niceMax="{$indicator.id === 'intensity' ? 5 : 10}"
       height="{['timing', 'intensity'].includes($indicator.id) ? 550 : 400}"
       on:showDownload="{loadDownload}"
       on:showShare="{loadShare}"
--- a/src/routes/tools/extreme-precipitation/_ExploreData.svelte
+++ b/src/routes/tools/extreme-precipitation/_ExploreData.svelte
@@ -107,7 +107,7 @@
     if (!$uncertaintyStore || $indicator.id !== "intensity") return "";
     const { lowSampleSize, nullCIValues } = $uncertaintyStore;
     if (lowSampleSize && nullCIValues) {
-      return removeTags(`${warningLowSampleSize}. ${warningMissingCI}`);
+      return removeTags(`${warningLowSampleSize} ${warningMissingCI}`);
     } else if (lowSampleSize) {
       return removeTags(warningLowSampleSize);
     } else if (nullCIValues) {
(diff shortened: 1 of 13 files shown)


src/routes/tools/extreme-precipitation/_store.js
```diff
@@ -9,6 +10,8 @@ import {
   DEFAULT_RETURN_PERIOD,
   DEFAULT_THRESHOLD_TYPE,
   DEFAULT_POLYGON_AGGREGATE_FUNCTION,
+  WARNING_LOW_SAMPLE_SIZE,
```
mukhtyar, line 13: remove unused constants

mukhtyar:

src/routes/tools/extreme-precipitation/_ExploreData.svelte
```diff
@@ -87,20 +91,41 @@
     dataByDate = groupDataByYear(flattenData(data));
   }
 
+  function removeTags(html) {
```
clhenrick, line 94: It's not too much to worry about but for the instances that this is being used it might just be more straightforward to save them as plain text files instead of markdown or just stash them as JS constants. 

I don't think we need to be super strict about using Markdown files for all content. If it's just a sentence there's not much gain (in terms of reducing the JS bundle size) from putting it in a Markdown file instead of having it as a JS constant.

src/routes/tools/extreme-precipitation/_ExploreData.svelte
```diff
@@ -87,20 +91,41 @@
...
-      return notificationText.replace(/<\/?[^>]+(>|$)/g, "");
+      return removeTags(notificationText);
+    }
+  }
+
+  function getUncertaintyText() {
+    if (!$uncertaintyStore || $indicator.id !== "intensity") return "";
+    const { lowSampleSize, nullCIValues } = $uncertaintyStore;
+    if (lowSampleSize && nullCIValues) {
+      return removeTags(`${warningLowSampleSize}. ${warningMissingCI}`);
```
clhenrick: I don't think you need the period here between the two `${}`

src/routes/tools/extreme-precipitation/_ExploreData.svelte
```diff
@@ -239,6 +265,7 @@
       units="{$climvar.units.imperial}"
       label="{indicatorLabel}"
       dataSource="{$titles.join(', ')}"
+      niceMax="{['intensity'].includes($indicator.id) ? 5 : 10}"
```
clhenrick: It would be simpler just to do: `$indicator.id === 'intensity' ? 5 : 10` right?

clhenrick (commented):
Just some minor suggestions / comments.

src/routes/tools/extreme-precipitation/_ExploreData.svelte
mukhtyar: good catch. fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/206/commits/609e284c54149dcec8b3d2aed9f4f5044ab9bcf6

src/routes/tools/extreme-precipitation/_ExploreData.svelte
mukhtyar: fixed in https://github.com/berkeley-gif/caladapt-website-2021/pull/206/commits/87be8c69e17888e595279e39a0a988f1c854e36a

PR #207, opened 2022-02-22
Bug fix chart axislabel spacing

mukhtyar:
This PR fixes the issue of overlapping y axis and tick value label for Line Area Charts ([trello issue](https://trello.com/c/sr2MwXfi)).


src/components/tools/Charts/LineArea/AxisY.svelte
```diff
@@ -9,13 +10,34 @@
...
     tickVals = $yScale.ticks().map(formatTick);
   }
   $: d = `M 0 0 L ${$width + $padding.left} 0`;
+
+  afterUpdate(() => {
+    // Use getComputedTextLength to get the computed length for the text within all the
+    // tick value labels. Use the maximum length value to update the padding for the axis label
+    // This prevent the axis label from overlapping the topmost y axis tick value.
+    try {
+      const tickLengths = Array.from(ref.querySelectorAll(".tick-label")).map(
```
clhenrick: This seems like it would work, but is there any reason you can't just use the first / last tick label? i.e. the one that's adjacent to the axis label? It could possibly end up looking weird if the longest label is at the bottom of the graph and not the top (not sure why that would happen but seems possible).

mukhtyar: Getting the last tick label makes sense. Updated in https://github.com/berkeley-gif/caladapt-website-2021/pull/207/commits/f54facc920c0876344845bffef1e0a2ff083bf5d

mukhtyar:
[Enhancement: Y axis title in LineArea chart runs into Y axis values](https://trello.com/c/sr2MwXfi/176-enhancement-y-axis-title-in-linearea-chart-runs-into-y-axis-values)

PR #213, opened 2022-02-25
Migrate lccs get wildfire data

mukhtyar:
This PR includes code for fetching `fire` data for the Area Burned Indicator. To accommodate the creation of ensemble & snapshot from individual models for `fire`, the data fetching code for other indicators has been refactored.

### DIFF
--- a/src/routes/tools/_common/helpers.js
+++ b/src/routes/tools/_common/helpers.js
@@ -26,11 +26,11 @@ export function buildEnvelope(_data) {
     (d) => d.date.getUTCFullYear()
   );
   return dataArr.map(([key, value]) => {
-    const sortedArr = sort(value);
+    const arrExtent = extent(value);
     return {
       date: new Date(Date.UTC(key, 0, 1)),
-      min: sortedArr[0],
-      max: sortedArr[1],
+      min: arrExtent[0],
+      max: arrExtent[1],
     };
   });
 }
--- a/src/routes/tools/_common/helpers.js
+++ b/src/routes/tools/_common/helpers.js
@@ -26,11 +26,11 @@ export function buildEnvelope(_data) {
     (d) => d.date.getUTCFullYear()
   );
   return dataArr.map(([key, value]) => {
-    const arrExtent = extent(value);
+    const [min, max] = extent(value);
     return {
       date: new Date(Date.UTC(key, 0, 1)),
-      min: arrExtent[0],
-      max: arrExtent[1],
+      min,
+      max,
     };
   });
 }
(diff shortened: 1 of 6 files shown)


src/routes/tools/_common/helpers.js
```diff
@@ -26,11 +26,11 @@ export function buildEnvelope(_data) {
     (d) => d.date.getUTCFullYear()
   );
   return dataArr.map(([key, value]) => {
-    const sortedArr = sort(value);
+    const arrExtent = extent(value);
```
mukhtyar: I made this change to create an envelope from more than 2 series. I did not end up using it, but left it in, because I thought this was a better implementation than the previous one.

src/routes/tools/local-climate-change-snapshot/_constants.js
```diff
@@ -57,7 +60,7 @@ export const OBSERVED = [
     color: COLOR_livneh.hex(),
   },
 ];
-export const SERIES = [
+export const SCENARIOS = [
```
mukhtyar, line 63: renamed these constants to make it more clear

mukhtyar:
[LCCS: Add data fetching and transformation](https://trello.com/c/ZwCSHzG2/171-lccs-add-data-fetching-and-transformation)

src/routes/tools/_common/helpers.js
clhenrick: Using `d3.extent` certainly makes it more clear what `value` is used for.

clhenrick: You could also destructure like 

```js
const [min, max] = extent(value);
```

then simplify the returned object below:

```js
return {
  date: new Date(...),
  min,
  max,
};
```

src/routes/tools/local-climate-change-snapshot/_constants.js
clhenrick: thank you!

src/routes/tools/local-climate-change-snapshot/_data.js
```diff
@@ -60,7 +46,7 @@ const fetchEvents = async ({
   if (error) {
     throw new Error(error.message);
   }
-  // Add slug to keep track of which timeseries the data comes from
+  // Add slug to keep track of which ensemble series the data comes from
```
clhenrick: Could you use a regex to get the slug?

clhenrick (commented):
Thanks @mukhtyar for getting all of this set up, it's a great start and is well documented and organized. I took a cursory look but will dig in more next week and the coming weeks. Appreciate your help with this before you leave!

(This is more of a note for myself since you likely won't be touching this code again but I might abstract the common parts of `getObserved` and `getProjections` into a separate function that could be then used in `getObserved` and `getProjections`)

PR #216, opened 2022-03-01
SLR CIS: add bookmark/share feature

clhenrick:
This adds a bookmark URL via a "share" button similar to other tools.

- [x] adds a `mapViewStore` that tracks the map's lat, lng, zoom, bbox
- [x] removes the `mapBBoxStore`
- [x] limits precision of lat, lng, zoom bookmark query params
- [x] adds helper fn for creating the bookmark query params
- [x] prevents title height from changing when toggling map layers
- [x] prevents missing data message from displaying when fetching raster metadata


clhenrick:
[SLR CIS: add share / bookmark feature](https://trello.com/c/PtyvUmJ8/175-slr-cis-add-share-bookmark-feature)

PR #218, opened 2022-03-04
SLR CIS: add map explainer text

clhenrick:
Adds the following map explainer text below the slippy map per the CEC request

> This map shows data coverage extents for the two models at lower zoom levels (e.g. at the county or state level resolutions), zoom in to view the coastal inundation datasets.


clhenrick:
[SLR CIS: update about & learn more content](https://trello.com/c/3ccVTJ66/180-slr-cis-update-about-learn-more-content)

PR #231, opened 2022-04-19
LCCS: add select location map

clhenrick:
Adds the interactive slippy map to the LCCS tool.

### DIFF
--- a/src/helpers/geocode.js
+++ b/src/helpers/geocode.js
@@ -13,18 +13,6 @@ export const caladaptGeocodingEndpoint = apiEndpoint;
 export const mapboxGeocodingEndpoint =
   "https://api.mapbox.com/geocoding/v5/mapbox.places";
 
-export const reverseGeocode = async (coords) => {
-  const url = `${mapboxGeocodingEndpoint}/${coords}.json`;
-  const geocodeParams = {
-    access_token: accessToken,
-  };
-  const [response, error] = await handleXHR(fetchData(url, geocodeParams));
-  if (error) {
-    throw new Error(error.message);
-  }
-  return response;
-};
-
 export const mapboxGeocodeParams = {
   country: "us",
   bbox: "-125,31,-113,44",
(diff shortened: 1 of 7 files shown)


clhenrick:
[LCCS: Implement Location Selection](https://trello.com/c/N43QXilN/165-lccs-implement-location-selection)

src/helpers/geocode.js
```diff
@@ -35,6 +23,11 @@ export const mapboxGeocodeParams = {
   access_token: accessToken,
 };
 
+/**
+ * geocode - uses MapBox Geocoding API to find geographic data for a street address or similar
+ * @param {string} searchStr - the street address, zipcode, locality, etc. to search for
+ * @returns {Object} GeoJSON FeatureCollection
+ */
 export const geocode = async (searchStr) => {
```
clhenrick: might as well pass the signal param here as well

src/routes/tools/local-climate-change-snapshot/_select-location/LocationMap.svelte
```diff
@@ -1,23 +1,95 @@
...
+    handleAbortFetch,
+    formatFeature,
+    formatLocaGridFeature,
+    reverseGeocodeBoundary,
+    reverseGeocodeAddress,
+  } from "./geocode-search";
 
-  export let boundaryType = "";
-  export let selectedLocation = null;
+  export let location;
```
clhenrick: name this `selectedLocation` for consistency with the `LocationForm`

src/routes/tools/local-climate-change-snapshot/_select-location/SelectLocation.svelte
```diff
@@ -5,18 +5,32 @@
...
+
+  function handleSelectLocation(value) {
+    locationStore.updateLocation(value);
+  }
+
+  function handleSelectBoundaryId(value) {
+    locationStore.updateBoundary(value);
+  }
+
+  function handleMapClick({ detail }) {
```
clhenrick, line 29: Should clicking on the map in an area where there is no location (e.g. the Pacific Ocean) clear the selected location and search value?

src/routes/tools/_common/stores.js
```diff
@@ -91,7 +91,6 @@ export const locationStore = (() => {
     subscribe,
     updateLocation: (location, isUpload = false) =>
       update((store) => {
-        if (!location) return;
```
clhenrick, line 94: Note to self: keep an eye out to see if this causes any bugs in other tools.

src/routes/tools/local-climate-change-snapshot/_select-location/SelectLocation.svelte
clhenrick: Inclined to say no, given that this may happen accidentally and be an annoying or confusing UX.

clhenrick:
@elehmer here is a Netlify preview deploy for you to test:
https://lccs-location-search-map--happy-engelbart-4042ee.netlify.app/tools/local-climate-change-snapshot

I've tried to maintain feature parity with the existing LCCS tool's location search form and map. The only thing that is different (other than some improved search form things I mentioned in the previous PR) is that when switching between boundary types other than address, the map won't automatically select a new boundary type given the previous boundary type's center. IMO this isn't a big deal because it's easy enough for someone to click on the map again, and the boundary type that gets automatically selected may not be what the user wants (e.g. going from city to census tract, or county to watershed, etc.). 

Let me know if you see anything odd or anything else that doesn't work similar as to how the existing LCCS location search works.

src/routes/tools/_common/stores.js
clhenrick: Also note to self: this caused a bug in the `locationStore` as any of the update methods should return the store, which wasn't happening here when `location` is falsey.

elehmer (approved):
LGTM! Great interaction of map and controls.

PR #232, opened 2022-04-21
LCCS tool: explore page header

clhenrick:
Adds the header area to the explore page which displays the selected location from the index page (location search).

## Tasks
- [x] handles keeping the location synced between index / select location page & explore data page
- [x] header title contains selected location title text
- [x] header contains static locator map of selected location
- [x] header contains button for generate pdf report
- [x] header contains button for select another location
- [x] clicking change location button navigates back to index page

## Other
- Updates the `StaticMap` shared component used by tools to be either a clickable button or a non-clickable `<div>` element. The LCCS tool has a separate "change location" button that is more obvious so we don't need the locator map to also serve as a change location mechanism.

- As a bonus this PR adds Sass Mixins for media queries that correspond to breakpoints in the Carbon Design Grid system. I plan on adding these into the main branch so they can be used elsewhere if needed before the LCCS tool migration is completed.

## Known Bugs
- Change Location in explore page will cause the index page to fail to load the location for some place names because their geographic center is located outside of their boundary polygon (e.g. San Francisco, Alameda). Problem has to do with using lng, lat, boundary type to reverse geocode to find a boundary polygon.

- Location title may not persist for certain geographies, e.g. a zipcode will return a street address when the page is refreshed or when clicking change location on explore. This is also due to how a store location from a lng, lat, boundary type is looked up via reverse geocoding.

- Locator static map in Explore fails to render if polygon geometry is invalid due to simplification. Larger issue with this component which could be addressed by rendering the static map using another strategy than MapBox's static map API.

- ~~The regex used for series slug in explore uses a "look behind" capturing group which [isn't yet supported in Safari](https://stackoverflow.com/a/51568859) so the LCCS will currently not work in Safari.~~ (fixed).

### DIFF
--- a/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
+++ b/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
@@ -12,15 +12,6 @@
   $: titleText = location
     ? `${defaultTitleText}:<br>${location.title}`
     : defaultTitleText;
-
-  // TODO: generate PDF from DOM
-  function noop() {}
-
-  async function changeLocation() {
-    // Note: using this method of page navigation as there appears to be a bug
-    // when using Sapper's goto method which doesn't correctly update the DOM.
-    window.location.pathname = "/tools/local-climate-change-snapshot/";
-  }
 </script>
 
 <style lang="scss">
@@ -68,14 +59,18 @@
       <h1>{@html titleText}</h1>
       {#if pageName === "explore"}
         <div class="header-btn-container">
-          <Button on:click="{noop}" size="field" icon="{Pdf16}"
-            >Generate Report</Button
+          <Button
+            on:click
+            size="field"
+            icon="{Pdf16}"
+            data-action="generate-report">Generate Report</Button
           >
           <Button
-            on:click="{changeLocation}"
+            on:click
             size="field"
             icon="{Location16}"
-            kind="tertiary">Change Location</Button
+            kind="tertiary"
+            data-action="change-location">Change Location</Button
           >
         </div>
       {/if}
--- a/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
+++ b/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
@@ -14,13 +14,14 @@
     ? `${defaultTitleText}:<br>${location.title}`
     : defaultTitleText;
 
-  function generateReport() {
-    // TODO...
-  }
+  // TODO: generate PDF from DOM
+  function noop() {}
 
-  function changeLocation() {
-    // TODO: set page scroll height to select location form & map.
-    goto("/tools/local-climate-change-snapshot/");
+  async function changeLocation() {
+    await goto("/tools/local-climate-change-snapshot/", { noscroll: true });
+    // give the page a little time to re-render before scrolling
+    await new Promise((resolve) => setTimeout(resolve, 350));
+    document.querySelector("#select-location").scrollIntoView();
   }
 </script>
 
@@ -70,7 +71,7 @@
         <h1>{@html titleText}</h1>
         {#if !showDescription}
           <div class="header-btn-container">
-            <Button on:click="{generateReport}" size="field" icon="{Pdf16}"
+            <Button on:click="{noop}" size="field" icon="{Pdf16}"
               >Generate Report</Button
             >
             <Button
--- a/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
+++ b/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
@@ -1,5 +1,6 @@
 <script>
   import { Grid, Row, Column } from "carbon-components-svelte";
+  import { StaticMap } from "~/components/tools/Location";
 
   export let iconPaths = [];
   export let showDescription = true;
@@ -39,6 +40,9 @@
     <Row>
       <Column lg="{10}" md="{8}" sm="{4}">
         <h1>{@html titleText}</h1>
+      </Column>
+      <Column lg="{{ span: 4, offset: 2 }}" md="{8}" sm="{4}">
+        <StaticMap height="{250}" location="{location}" useButton="{false}" />
       </Column>
     </Row>
     {#if showDescription}
--- a/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
+++ b/src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
@@ -5,8 +5,8 @@
   import { StaticMap } from "~/components/tools/Location";
 
   export let iconPaths = [];
-  export let showDescription = true;
-  export let location;
+  export let location = null;
+  export let pageName;
 
   const defaultTitleText = "Local Climate Change Snapshot";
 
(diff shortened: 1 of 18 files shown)


src/routes/tools/local-climate-change-snapshot/_common/Header.svelte
```diff
@@ -30,8 +67,26 @@
       </Column>
     </Row>
     <Row>
-      <Column>
-        <h1>Local Climate Change Snapshot</h1>
+      <Column lg="{10}" md="{8}" sm="{4}">
+        <h1>{@html titleText}</h1>
+        {#if !showDescription}
```
clhenrick: Maybe rename `showDescription` to something else like `contentOnly`? Or use the route name e.g. 

```svelte
{#if page === "explore"}
```

so that it's more obvious why something is or isn't rendered.

src/routes/tools/local-climate-change-snapshot/_explore/ExploreData.svelte
```diff
@@ -9,7 +7,6 @@
   import CategoryResources from "./CategoryResources.svelte";
 </script>
 
-<LocationTitle />
```
clhenrick, line 12: Should remove this component

clhenrick:
[LCCS: Header component](https://trello.com/c/D8B8RoMt/197-lccs-header-component)

src/routes/tools/local-climate-change-snapshot/explore.svelte
```diff
@@ -88,6 +92,16 @@
   $: console.log("chart data", $chartDataStore);
   $: console.log("snapshot data", $snapshotDataStore);
 
+  function handleHeaderBtnClick(event) {
+    if (event.target.dataset.action === "change-location") {
+      // Note: using this method of page navigation as there appears to be a bug
```
clhenrick: change to `// Note we're using...` and add `this may have to do with explore.svelte fetching index.json.js`

src/routes/tools/local-climate-change-snapshot/explore.svelte
```diff
@@ -173,19 +187,41 @@
...
     indicatorListStore.set(indicators);
     indicatorStore.set(indicators.find((d) => d.id === indicator));
     unitsStore.set({ imperial });
-    const loc = await setInitialLocation(+lng, +lat, boundary);
+  }
+
+  async function setLocationStoreOnLoad(lat, lng, boundaryType) {
+    let loc = DEFAULT_LOCATION;
+    if (
+      isValidNumber(lat) &&
```
clhenrick: maybe could wrap this in an `isValidLocationParams` function:

```js
function isValidLocationParams(lng, lat, boundaryType) {
  return isValidNumber(lng) && isValidNumber(lat) && VALID_BOUNDARY_TYPES.has(boundaryType)
}
```
so that it can be shared between the index and explore pages.

src/routes/tools/local-climate-change-snapshot/explore.svelte
```diff
@@ -173,19 +187,41 @@
...
+      await setLocationStoreOnLoad(+lat, +lng, boundaryType);
+    }
     categoryListStore.set(categories);
     indicatorListStore.set(indicators);
     indicatorStore.set(indicators.find((d) => d.id === indicator));
     unitsStore.set({ imperial });
-    const loc = await setInitialLocation(+lng, +lat, boundary);
+  }
+
+  async function setLocationStoreOnLoad(lat, lng, boundaryType) {
```
clhenrick: could abstract this function so that it is shared between the index and explore pages, or tool pages in general.

src/routes/tools/local-climate-change-snapshot/index.svelte
```diff
@@ -34,12 +34,86 @@
...
+    } = $page;
+    lat = +lat;
+    lng = +lng;
+    if (
+      isValidNumber(lat) &&
+      isValidNumber(lng) &&
+      VALID_BOUNDARY_TYPES.has(boundaryType)
+    ) {
+      try {
+        loc = await setInitialLocation(lng, lat, boundaryType);
```
clhenrick: Would not want to fallback to the default location here, it should be blank for users when they first arrive at the page.

src/routes/tools/local-climate-change-snapshot/index.svelte
```diff
@@ -34,12 +34,86 @@
...
+      } catch (error) {
+        console.warn(error);
+      }
+    }
+    if (loc) {
+      // TODO: if boundaryType is locagrid, make selectedLocation.geometry a point
```
clhenrick: Address this before merging

src/routes/tools/local-climate-change-snapshot/explore.svelte
```diff
@@ -173,19 +188,30 @@
...
+    } = getInitialConfig(query, DEFAULT_INITIAL_CONFIG);
+    lat = +lat;
+    lng = +lng;
     // Set intial values for stores
+    if (!$location || !$boundary) {
+      let loc;
+      try {
+        loc = await getLocationFromQuery(lng, lat, boundaryType, true);
+      } catch (error) {
+        console.warn(error);
```
clhenrick: Should maybe display a warning here via the notify as if the location isn't correctly set the app won't load.

src/routes/tools/local-climate-change-snapshot/_helpers.js
```diff
@@ -1,17 +1,53 @@
...
+    isValidNumber(lng) &&
+    VALID_BOUNDARY_TYPES.has(boundaryType)
+  );
+};
+
+export const getLocationFromQuery = async (
+  lng,
+  lat,
+  boundaryType,
+  useDefaultLocation = false
```
clhenrick: could make this default param set to `DEFAULT_LOCATION` and then make the function caller deliberately set it to `false` which would eliminate the need for the `fallbackLocation` variable within the function.

src/routes/tools/local-climate-change-snapshot/explore.svelte
```diff
@@ -173,19 +184,39 @@
...
+      lng,
+      boundary: boundaryType,
+      indicator,
+      imperial,
+    } = getInitialConfig(query, DEFAULT_INITIAL_CONFIG);
+    lat = +lat;
+    lng = +lng;
     // Set intial values for stores
+    if (!$location || !$boundary) {
+      let loc;
```
clhenrick: could maybe wrap this all in a separate function

src/routes/tools/local-climate-change-snapshot/index.svelte
```diff
@@ -34,12 +34,82 @@
...
+        fallbackLocation: undefined,
+      });
+    } catch (error) {
+      console.log(error);
+    }
+    if (loc) {
+      selectedLocation = loc;
+      searchValue = loc.title;
+      // TODO: override when fallback geom is present.
+      selectedRadio = boundaryType;
```
clhenrick, line 85: Address this maybe by doing

```js
selectedRadio = selectedLocation === DEFAULT_LOCATION ? "locagrid" : boundaryType;
```
but `DEFAULT_LOCATION` has its geometry modified to be a point so might want to check against `searchValue` instead, e.g.

```js
selectedRadio = searchValue === DEFAULT_LOCATION.title ? "locagrid" : boundaryType;
```

clhenrick: Actually this isn't necessary or desirable as we want the index page to load with no location selected if a valid one is not found via the URL query params.

clhenrick:
@elehmer here is a [Netlify preview deploy](https://lccs-explore-header--happy-engelbart-4042ee.netlify.app/tools/local-climate-change-snapshot) for you to try out. The PR description up top describes what this PR should accomplish and also lists known bugs. Let me know if you discover anything else that doesn't seem right. Thanks!

elehmer (approved):
LGTM. Pretty large rewrite. Nicely done!

PR #233, opened 2022-04-26
use sass media query mixins

clhenrick:
- [x] adds CSS media query Sass mixins via the [include-media](https://www.npmjs.com/package/include-media) scss library.
- [x] replaces existing media queries with their corresponding mixin to encourage consistency

### DIFF
--- a/src/components/tools/Partials/Dashboard/index.svelte
+++ b/src/components/tools/Partials/Dashboard/index.svelte
@@ -82,7 +82,7 @@
     }
   }
 
-  @include mobile-tablet {
+  @include mq-mobile-tablet {
     .dashboard {
       .content {
         padding-right: 0;
--- a/src/components/tools/Partials/Dashboard/index.svelte
+++ b/src/components/tools/Partials/Dashboard/index.svelte
@@ -22,6 +22,8 @@
 </script>
 
 <style lang="scss">
+  @import "scss/site/mixins/media-queries";
+
   .dashboard {
     display: flex;
     max-width: 99rem;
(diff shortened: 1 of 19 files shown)


src/components/tools/Partials/Dashboard/index.svelte
```diff
@@ -61,23 +63,26 @@
   }
   // end overrides for tab content area
 
-  @media (max-width: 1000px) {
+  @include media("<=1000px") {
```
clhenrick: use `@include custom-1000` here

src/components/tools/Partials/Dashboard/index.svelte
```diff
@@ -61,23 +63,26 @@
...
 
       .sidebar {
-        width: 100%;
         padding-left: 2rem;
       }
     }
   }
 
-  @media (max-width: 672px) {
+  @include mobile-tablet {
```
clhenrick: Double check this, might be `@include media("<=medium") {`?

src/scss/site/mixins/_media-queries.scss
```diff
@@ -0,0 +1,50 @@
...
+
+// use when applying media queries for mobile device width viewports
+@mixin mobile {
+  @include media("<=small") {
+    @content;
+  }
+}
+
+// custom media query for 1000px
+@mixin custom-1000 {
```
clhenrick: Maybe name this something like `mq-custom-1000px` instead so that it's clear what it is? Could also prefix the other media query mixins with `mq-` for clarity as well.

clhenrick:
@elehmer this is mostly cleaning up the use of [CSS @media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/@media) in the code. We mainly want to make sure nothing in the layout is broken. I added this PR because I'm using this technique in the LCCS tool migration and it helps keep use of media queries more consistent in the code.

- [Netlify deploy preview](https://use-include-media-sass--happy-engelbart-4042ee.netlify.app)

elehmer (approved):
LGTM. Layout appears unaffected at least on my desktop.

PR #234, opened 2022-04-27
bug fix: location store

clhenrick:
Removes an early return when updating the `location` property of the `locationStore` (Svelte store) to prevent the store from breaking.

This doesn't fix the issue I [recently mentioned in Trello](https://trello.com/c/8JmopK9Q/198-bug-bookmarked-location-may-fail-to-load-desired-location-for-any-tool), but it is related and may cause problems elsewhere.


elehmer (approved):
I was able to change location on 4-5 tools on both Ubuntu Chromium/Firefox and Windows 10 Edge/Firefox

PR #237, opened 2022-05-09
StaticMap component: feature & bug fix

clhenrick:
Updates to the `StaticMap` component:
- Enables wrapping the locator map image in either a button or div (the latter for preventing interaction, for example with use in the LCCS tool "explore" page).
- Fixes the loading, loaded, and error states so that the error message will display when rendering the map fails (e.g. for boundaries like Tehama County) and that the loading message will display when the location is changed.

**NOTE:** this PR makes some temporary adjustments to this component but does not fix [the issue](https://trello.com/c/cKnxzgSN) which causes it to occasionally not render.

### DIFF
--- a/src/components/tools/Location/StaticMap.svelte
+++ b/src/components/tools/Location/StaticMap.svelte
@@ -66,6 +66,7 @@
     )}`;
   }
 
+  // FIXME: this creates invalid geometries for some boundaries, e.g. Tehama County
   function createOverlay(geojson, tolerance = 0.005) {
     const overlay = encodeURIComponent(JSON.stringify(geojson));
     if (overlay.length < 7500) {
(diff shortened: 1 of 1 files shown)


clhenrick:
@elehmer here is a [netlify preview deploy](https://static-map-updates--happy-engelbart-4042ee.netlify.app/tools/) for you to test out with this PR.

clhenrick:
@elehmer for an example of a boundary that does not render, try changing the location to Tehama County. An error message should display.

clhenrick:
[Bug: tool locator map fails to load due to invalid polygon](https://trello.com/c/cKnxzgSN/196-bug-tool-locator-map-fails-to-load-due-to-invalid-polygon)

elehmer:
This is a much more graceful handling of the invalid polygon in the static map. Seems to work great in several tools that I tried out.

PR #252, opened 2022-12-06
Reformat LOCA grid cell title

elehmer:
This PR adapts the website to the change in the LOCA grid cell title format from:

`LOCA Grid Cell %longitude%, %latitude%` to `LOCA Grid Cell %latitude%, %longitude%`

### DIFF
--- a/src/helpers/geocode.js
+++ b/src/helpers/geocode.js
@@ -13,6 +13,9 @@ export const caladaptGeocodingEndpoint = apiEndpoint;
 export const mapboxGeocodingEndpoint =
   "https://api.mapbox.com/geocoding/v5/mapbox.places";
 
+// Cache busting variable. Increment to get refreshed content from API. Last increment: 2022-12-06
+const v = 2;
+
 export const mapboxGeocodeParams = {
   country: "us",
   bbox: "-125,31,-113,44",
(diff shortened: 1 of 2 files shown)


src/helpers/geocode.js
```diff
@@ -68,7 +68,7 @@ export const searchBoundaryLayer = async (searchStr, boundaryId) => {
   const searchParams = {
     srs: 4326,
     search: searchStr,
-    v: 1, // cache bust 2022-05-09
```
mukhtyar, line 71: An optional update is to create a new `const v = 2` at the top of this file and then use it for three other places it appears.

mukhtyar (approved):

src/helpers/geocode.js
elehmer: > An optional update is to create a new `const v = 2` at the top of this file and then use it for three other places it appears.

True, that would be more efficient wouldn't it. I'll make the changes...

elehmer:
Better?

src/helpers/geocode.js
elehmer: Better?

elehmer:
[Add hemisphere notation to lat/long display for LOCA grid cells](https://trello.com/c/R2ik2vrt/212-add-hemisphere-notation-to-lat-long-display-for-loca-grid-cells)

PR #259, opened 2023-03-16
Update blog text

mukhtyar:


mukhtyar:
Merging it without review because the updates are minor

PR #265, opened 2023-03-27
Increase roadmap size

mukhtyar:
This PR wraps the roadmap graphic as a link, so users can click on it to view a larger version of the image in a new tab.


melijimenez11 (approved):
Looks good

PR #269, opened 2023-03-30
Add info static files dir

mukhtyar:
This PR describes a new static `files` dir on the server that is used to host word documents, pdf files, etc. (any file resource that CEC wants us to host), but is not tracked in the github repo.

### DIFF
--- a/src/routes/index.svelte
+++ b/src/routes/index.svelte
@@ -235,7 +235,7 @@
       <div class="bx--row">
         <div class="bx--col-lg-12">
           <h3>
-            Discover how our <a href="/grants" target="_blank"
+            Discover how our <a href="/grants"
               >data development grants and research projects</a
             >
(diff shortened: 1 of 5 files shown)


src/routes/index.svelte
```diff
@@ -235,7 +235,7 @@
       <div class="bx--row">
         <div class="bx--col-lg-12">
           <h3>
-            Discover how our <a href="/grants" target="_blank"
+            Discover how our <a href="/grants"
```
mukhtyar, line 238: @melijimenez11 For some reason I no longer see the scrolling bug, so I removed the `target=_blank`.

melijimenez11: Sounds good

melijimenez11 (approved):
All looks good to me

